PR #139, opened 2020-11-10
Stdp fix

sky1204:
Creating a pull request for CUDA since this is where I forked my branch, and CUDA hasn't been merged with master. These are just the fixes made to correct the indexing of the circular queue and synapse index map.


stiber (dismissed):
Looks good. There are a lot of changes on master, but I think the only potential conflicts will be tweaks of the .gitignore that should cause many of these files to not be tracked by git.

stiber:
@rashwini21 this pull request may have useful CUDA code or it may not. Just want to make sure before we toss this and its associated branch as having already been (manually) merged into the codebase. Snigdha starts the pull request off by mentioning CUDA and fixes to event queueing and the synapse index map, so I think that looking through this carefully will allow us to either be more confident in the master branch CUDA code or catch bugs in it.

PR #153, opened 2021-03-18
CMake out of source build

christopherdokeefe:
CMake is now configured to build out of the source files. Rather than building in the root directory, go into the `build` directory and type `cmake ..` followed by `make`. This will create all the build files, libraries and executables inside the `build` directory rather than in the source files.

NOTE: When running the simulation you must use `../` before the configuration file. For example 

`./braingrid -c ../configfiles/test-tiny.xml`

- Created README file inside of build with instructions on how to build.
- Changed all paths in files to include a `../` at the beginning since the path to files is now changed.
- Updated top-level CMakeLists.txt that includes more information about the build.
- All the config file paths need to be updated before the merge, which will be done after the changes are approved.

Edit: Updated the GitHub actions to run in the build directory as well.


stiber (changes requested):
Looks good. Needs two changes, and I also have two notes. First the changes: there are two library files under `lib` that need to be removed from the repo. They're probably already in the `.gitignore`. Likely just need to do `git rm --cached FILENAME`.

OK, then the notes: Likely, the two config files will have conflicts with the most recent commits to master that include generalization of nomenclature. Should be obvious when you go to merge.

Simulator/Edges/AllEdges.h
```diff
@@ -32,7 +32,7 @@
 #include "IAllEdges.h"
 
 // cereal
-#include <ThirdParty/cereal/types/vector.hpp>
+#include <../ThirdParty/cereal/types/vector.hpp>
```
stiber, line 35: Why using angle brackets, instead of quotes?

stiber (commented):
One question. Let me know if previous review comment settled.

Simulator/Edges/AllEdges.h
christopherdokeefe: Yeah, you're right. The brackets should only be used when locating library files outside of the project. I will change that.

christopherdokeefe:
- The lib files are deleted in this branch so the files will be gone completely when the branch is merged. 
- I also fixed the small syntax problem with the include path. 
- I haven't changed the conflicting config files yet because the changes I made are pretty small and I'm not familiar with changes that happened in master. Once this branch is merged I will go through the config files and make the additional changes needed. If that's okay, I am ready to merge this branch.

stiber (approved):
OK, this is good to go. Tori and Vivek know about the changes for 911, going forward.

Issue #147, opened 2021-02-12
Change NULL to nullptr


stiber:
There are still some places where this very minor cleanup is needed.

BodilessSleeper:
Does this relate to all files (eg files in ThirdParty) and even comments (eg in Simulator/Core/Simulator.cpp)?
`/* PInput not in project yet
 if (pInput != NULL) {
     cout << "Initializing input." << endl;
     pInput->init();
 }
 */`

stiber:
Needs to be fixed after subsequent merges of branches with large changes.

PR #162, opened 2021-04-10
Changed all occurrences of NULL to nullptr

BodilessSleeper:
This tackles issue #147 (the only problematic file could be Simulator/Utils/Timer.h which had NULL in a variable name where one could argue it's easier to have it left as it was while on the other hand one could say it's a bit confusing to have a variable that has NULL in it but actually references nullptr although I don't see this variable name or even the variable in place ever being used somewhere in the code)

### DIFF
--- a/Simulator/Connections/ConnectionsFactory.cpp
+++ b/Simulator/Connections/ConnectionsFactory.cpp
@@ -47,5 +47,5 @@ Connections *ConnectionsFactory::invokeCreateFunction(const string &className) {
       if (className == i->first)
          return i->second();
    }
-   return NULL;
+   return nullptr;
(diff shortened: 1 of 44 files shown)


stiber:
> This tackles issue #147 (the only problematic file could be Simulator/Utils/Timer.h which had NULL in a variable name where one could argue it's easier to have it left as it was while on the other hand one could say it's a bit confusing to have a variable that has NULL in it but actually references nullptr although I don't see this variable name or even the variable in place ever being used somewhere in the code)

That code is there to provide a substitute for `gettimeofday()` for Win32 platforms; it's called by code in Timer.cpp.

Issue #157, opened 2021-03-30
Fix misspelled filename


stiber:
Not just in the comment; also the filename itself

https://github.com/UWB-Biocomputing/Graphitti/blob/2d15fe3ae34c3f4900946179d3c6f3f2b7780908/Testing/Utils/ParamaterManagerTests.cpp#L2

torisalvatore:
done in chris's commit.

PR #163, opened 2021-04-12
Fixed a typo in the filename and the comment in the same file

BodilessSleeper:
Resolves issue #157


stiber:
Seems that an internal project member had fixed this just a little while ago; he just hadn't closed the issue. We are still a work in progress, and are learning to be more disciplined in doing things like linking pull requests to issues and/or indicating in commits that they close particular issues.

PR #168, opened 2021-04-13
Changed all occurences of NULL to nullptr

BodilessSleeper:
Fixes issue #147


stiber:
We have just gone through major merges of large changes in the code base, so this will have to be reviewed very carefully. Files have moved, files have been renamed, classes have been merged/split/created, and class and variable names have been changed.

Simulator/Core/BGDriver.cpp
```diff
@@ -137,7 +137,7 @@ int main(int argc, char *argv[]) {
 
    // INPUT OBJECTS ARENT IN PROJECT YET
    // Terminate the stimulus input
-//   if (pInput != NULL)
+//   if (pInput != nullptr)
```
stiber, line 140: File name has changed.

BodilessSleeper: Eh, you could've merged this PR first and then #170 and we would've had no problems, but I guess I'll just redo the problematic files (not sure if this is the only one of all the changed files)

BodilessSleeper:
Should be done now

stiber (approved):

Issue #164, opened 2021-04-13
setup methods rename to allocateMemory()


torisalvatore:
all setup methods just allocate memory. this should be reflected in the name of the function to promote clarity for the newcomers.

VivekGandhi54:
There are setup methods for a variety of classes, should they all be called "allocateMemory", or keep the class descriptions?

setupEdges() --> allocateMemory() and setupVertices -->allocateMemory()
OR
setupEdges() --> allocateEdgeMemory() and setupVertices -->allocateVertexMemory()?

stiber:
I would prefer a generic name that doesn't include the class name, so that we know we can just look for an `allocateMemory()` method in any class we run into. This would apply even to classes that allocate memory in their constructors — instead, have the constructors call a private or protected `allocateMemory()` method.

PR #173, opened 2021-04-16
Renamed all setup methods

VivekGandhi54:
Fixes issue #164 - renames setupConnection, setupEdges, etc. to allocateMemory()
Moves edgSign from AllEdges to AllNeuroEdges

### DIFF
--- a/Simulator/Connections/Neuro/ConnStatic.cpp
+++ b/Simulator/Connections/Neuro/ConnStatic.cpp
@@ -43,14 +43,14 @@ ConnStatic::~ConnStatic() {
 ///  @param  edges  The Synapse list to search from.
 void ConnStatic::setupConnections(Layout *layout, IAllVertices *vertices, AllEdges *edges) {
    int numVertices = Simulator::getInstance().getTotalVertices();
-   vector<DistDestVertex> distDestVertices[numVertices];
+   vector<DistDestVertex> distDestVertices;
 
    int added = 0;
 
    LOG4CPLUS_INFO(fileLogger_, "Initializing connections");
 
    for (int srcVertex = 0; srcVertex < numVertices; srcVertex++) {
-      distDestVertices[srcVertex].clear();
+      distDestVertices.clear();
 
       // pick the connections shorter than threshConnsRadius
       for (int destVertex = 0; destVertex < numVertices; destVertex++) {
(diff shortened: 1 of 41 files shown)


Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -41,16 +41,16 @@ ConnStatic::~ConnStatic() {
 ///  @param  layout    Layout information of the neural network.
 ///  @param  vertices   The Vertex list to search from.
 ///  @param  edges  The Synapse list to search from.
-void ConnStatic::setupConnections(Layout *layout, IAllVertices *vertices, AllEdges *edges) {
+void ConnStatic::allocateMemory(Layout *layout, IAllVertices *vertices, AllEdges *edges) {
```
stiber, line 44: Does this method do more than allocate memory?

Simulator/Core/GPUModel.cu
```diff
@@ -82,13 +82,13 @@ void GPUModel::deleteDeviceStruct(void** allVerticesDevice, void** allEdgesDevic
 }
 
 /// Sets up the Simulation.
-void GPUModel::setupSim()
+void GPUModel::allocateMemory()
```
stiber, line 85: Seems like this also does more than allocate memory.

Simulator/Core/Model.cpp
```diff
@@ -68,17 +68,17 @@ void Model::createAllVertices() {
 }
 
 /// Sets up the Simulation.
-void Model::setupSim() {
+void Model::allocateMemory() {
```
stiber, line 71: Seems that this does more than call memory allocation methods. At least, at the end, it creates the index map. That might be just a memory allocation method, however.

VivekGandhi54: Yeah, you're right. There's some inconsistent usage of the setup methods, probably why they weren't called allocateMemory to begin with.
Should I undo all the renaming, or just keep the changes for setupEdges and setupVertices?

Simulator/Layouts/Layout.cpp
```diff
@@ -66,7 +66,7 @@ shared_ptr<IAllVertices> Layout::getVertices() const {
 
 /// Setup the internal structure of the class.
 /// Allocate memories to store all layout state, no sequential dependency in this method
-void Layout::setupLayout() {
+void Layout::allocateMemory() {
```
stiber, line 69: I think this is an example where it's not pure memory allocation...

PR #184, opened 2021-04-30
Tori vivek 911 math jc

torisalvatore:
progress on 911 code, doesn't break on GPU version


stiber (changes requested):
Almost ready; see below for requested changes.

stiber (approved):

Issue #199, opened 2021-05-16
Why does AllEdges::createEdgeIndexMap() have dynamically allocated arrays of vectors?


stiber:

stiber:
And, for that matter, which is `EdgeIndexMap` dynamically allocated?

VivekGandhi54:
Same as #188

Issue #188, opened 2021-05-06
Fix synapseIndexMap_ creation and deletion


VivekGandhi54:
An `EdgeIndexMap` object is created and assigned to `Connections::synapseIndexMap_` at the end of every epoch and at the start of the simulation. ~~At the moment, this causes a memory leak between epochs.~~[shared_ptr destroys it when it's no longer used] It should be created in `Connections::setupConnections` and overwritten by `createEdgeIndexMap()` with each call.

~~The size of the map doesn't change~~, and `incomingEdgeCount_` keeps track of the indices used anyway.
~~Destroy `synapseIndexMap_` in the Connections destructor.~~ [shared_ptr destroys it when it's no longer used]

VivekGandhi54:
The size of the map could be `numVertices * maxEdgesPerVertex`. Then we wouldn't have to dynamically allocate it, but there would be some redundancy.

stiber:
I think we will have a decision soon (by end of spring 2021) regarding how we will replace the mishmash of vectors and dynamically allocated arrays. At that point, we should move to complete encapsulation of such things in a single type (maybe `valarray`), with fixed size allocation for the entire simulation duration.

This is important, because that is how things are allocated on the GPU and our approach is to have the CPU code mirror the GPU code.

PR #202, opened 2021-05-19
EdgeIndexMap Fix

VivekGandhi54:
Proposed fix for EdgeIndexMap creation, addresses issues #188, #199

Does not switch to `valarrays` or `vectors`, that can be done independently later.
Allocates some more memory than it used to, but shouldn't affect how it works.
It used to allocate the minimum required space, now allocates space for `numVertices * maxEdgesPerVertex` edges.


stiber:
I don't have a problem with this,  but right now the simulator isn't working. I would prefer that we hold off on any additional tweaks/modifications to the simulator until we have it completely operational.

stiber (approved):

PR #212, opened 2021-07-05
Fix parameter manager

### LINKED ISSUE #208: ParameterManager error messages produced during at least some simulator runs
We get the following error messages when running `test-tiny.xml` and `test-small.xml`. Might be true for other config files:

```
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
Parsed parameter could not be parsed as a float. Value: 
```

jbrown725:
closes #208


Simulator/Utils/ParameterManager.cpp
```diff
@@ -100,13 +100,17 @@ bool ParameterManager::checkDocumentStatus() {
...
    if (!checkDocumentStatus()) return false;
-   if (!TinyXPath::o_xpath_string(root_, xpath.c_str(), xpath)) {
+
+   // temp string holds evaluation of value at xpath
+   string temp;
+   // raise error if tinyxml cannot compute the xpath's value or returns empty
+   if (!TinyXPath::o_xpath_string(root_, xpath.c_str(), temp) || temp == "") {
       cerr << "Failed loading simulation parameter for xpath "
            << xpath << endl;
       // TODO: possibly get better error information?
```
stiber, line 110: Is this possible?

configfiles/test-large-connected.xml
```diff
@@ -80,5 +92,10 @@
         <probedNListFileName name="prbNListFileName" type="InputFile"/>
       </LayoutFiles>
     </LayoutParams>
+
+    <RecorderParams class="XmlGrowthRecorder" name="RecorderParams">
+      <probedNListFileName name="prbNListFileName" type="InputFile">../configfiles/NList/probedNList_900.xml</probedNListFileName>
```
stiber, line 97: I think this file is for a 900-neuron (30x30) simulation. We can talk about this at our upcoming meeting. I'll not comment on the other config files.

stiber (commented):
I think there's just two matters remaining for us to discuss in this:

1. Double-check some things in the config files (I think just the probed neuron list files, but maybe not).
2. What exactly are the PM error messages that these mods are producing and do they include as much information as possible (I believe it's file line and column number)? This is mostly spurred by a "TODO" comment still in the code.

stiber (approved):

PR #227, opened 2021-07-21
ConnGrowth fix

jbrown725:
`ConnGrowth` has a variable `growthParams_.maxRate` which was not being initialized. This variable had a default value of 2.35099e-38 which was being used in later calculations for connection growth. After initializing this variable, Graphitti now outputs identical results to BrainGrid in all the tests I've run.


jbrown725:
A check for epsilon having a value of 0 is done to avoid division by 0. It currently just writes a warning to the log, but this could change to a console warning or a termination of the simulator if need be

Simulator/Connections/Neuro/ConnGrowth.cpp
```diff
@@ -107,6 +107,13 @@ void ConnGrowth::loadParameters() {
    ParameterManager::getInstance().getBGFloatByXpath("//GrowthParams/targetRate/text()", growthParams_.targetRate);
    ParameterManager::getInstance().getBGFloatByXpath("//GrowthParams/minRadius/text()", growthParams_.minRadius);
    ParameterManager::getInstance().getBGFloatByXpath("//GrowthParams/startRadius/text()", growthParams_.startRadius);
+
+   // initial maximum firing rate
+   if (growthParams_.epsilon != 0) {
+      growthParams_.maxRate = growthParams_.targetRate / growthParams_.epsilon;
+	} else {
+      LOG4CPLUS_WARN(fileLogger_, "Parameter GrowthParams/epsilon/ has a value of 0" << endl);
```
stiber, line 115: I think this should be a FATAL level log entry, with an exit() afterwards.

stiber (changes requested):
Just one tiny mod. Looks great!

PR #229, opened 2021-07-22
Various small issues

VivekGandhi54:


CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ set(CMAKE_CXX_STANDARD 11)
 ############################################################################################
 set(ENABLE_CUDA NO)
 
+# set(DEBUG_MODE YES) for debugging, no optimizaiton
```
stiber, line 17: Typo

CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ set(CMAKE_CXX_STANDARD 11)
...
+# set(DEBUG_MODE NO) for production code, -O3 optimization enabled
+set(DEBUG_MODE NO)
+
 # Enables the CUDA langauge and locates the NVCC compiler.
```
stiber, line 21: Typo

Simulator/Core/Driver.cpp
```diff
@@ -68,6 +71,16 @@ void serializeSynapses();
...
+
+   // Doesn't work if placed in parseCommandLine because -c is a required flag,
+   // and ParamContainer throws an error
+   if (argc == 2) {
+      if (strcmp(argv[1], "-v") == 0 || strcmp(argv[1], "--version") == 0){
+         cout << "Git commit ID: " << GIT_COMMIT_ID << endl;
+         exit(0);
+      }
+   }
+
```
stiber, line 83: Let's fix this in our usage of ParamContainer, rather than adding separate code just for this.

stiber (changes requested):
Overall, looks good. Now that we have the `-v` option, then `-c` is no longer always a required command line argument. We just need to adjust this so that we can use ParamContainer; we also need to make sure that a config file is provided is there is no `-v` option.

I'll consult with the Workbench team about whether to use the short or long commit ID and get back to you.

Simulator/Core/Driver.cpp
VivekGandhi54: If it executes in `parseCommandLine`, then this gets printed to the terminal first:

```[TRACE][2021-07-22 22:28:43] Instantiating Simulator
[TRACE][2021-07-22 22:28:43] Parsing command line```

**Comment by VivekGandhi54** [comment]
If it executes in `parseCommandLine`, then this gets printed to the terminal first, which may be undesirable:
```
[TRACE][2021-07-22 22:28:43] Instantiating Simulator
[TRACE][2021-07-22 22:28:43] Parsing command line
```

**Comment by stiber** [comment]
> If it executes in `parseCommandLine`, then this gets printed to the terminal first, which may be undesirable:
> 
> ```
> [TRACE][2021-07-22 22:28:43] Instantiating Simulator
> [TRACE][2021-07-22 22:28:43] Parsing command line
> ```

Let's talk about this at our next meeting. Will be faster than comments here; is a simple thing.

File: CMakeLists.txt [hunk 4]
```diff
@@ -40,6 +44,23 @@ else()
         message("-- HDF5 library was not located. Please only use XML recorders.")
 endif()
 
+# Get short git commit ID
+# Change to "git rev-parse HEAD" for full commit ID
+execute_process(
+        COMMAND
+        git rev-parse --short HEAD
```
stiber, line 51: Yup, change it to full ID. This is great!

stiber (commented):
Looking through `AllEdges::createEdgeIndexMap()`, I have to say that there are a variety of questions I have about it. Given that the change to it may be minor, I don't think we necessarily should block this pull request. But it would be good to talk through it when we have the time.

Simulator/Core/Driver.cpp
```diff
@@ -207,7 +212,10 @@ bool parseCommandLine(int argc, char *argv[]) {
        || (cl.addParam("deserializefile", 'r', ParamContainer::filename,
                        "simulation deserialization filepath (enables deserialization)") != ParamContainer::errOk)
        || (cl.addParam("serializefile", 'w', ParamContainer::filename,
-                       "simulation serialization filepath (enables serialization)") != ParamContainer::errOk)) {
+                       "simulation serialization filepath (enables serialization)") != ParamContainer::errOk)
+       || (cl.addParam("version", 'v', ParamContainer::novalue,
+                       "current git commit ID") != ParamContainer::errOk)) {
```
stiber, line 217: I hate to do this, but the docstring should be "output current git commit ID and exit".

stiber (changes requested):
Just one teeny change...

stiber (approved):
Yay!

PR #231, opened 2021-07-26
RNG overhaul

VivekGandhi54:


VivekGandhi54:
Changes:

- Explicitly declared RNG as a shared library in Cmake, which fixes the GPU linking error.
- Renamed "MersenneTwister.h" -> "MTRand.h".
- Cleaned up some bad code in `MTRand` and `Norm`.
- Created `RNGFactory`, which creates instances of `MTRand` or `Norm`.
- Created Norm re-seeding methods, with same behavior as initializers.
- Replaced the vector that contained only one instance of `Norm` (`rgNormrnd`) with a single instance of `MTRand`.
- `rgNormrnd` can be initialized to a `Norm` or `MTRand` object from the config files.
- Both rng objects (`rgNormrnd` and `rng`) can now be seeded individually from the config files.

VivekGandhi54:
Some comments, issues, and questions:
1. I haven't made the changes in all config files yet.
2. The initialization and seeding assumes that `Norm` assumes that `m = 0.0`, and `s = 1.0`. Is this an okay assumption to make? Should I create rng `loadParameters()` functions that specifies these?
3. I need feedback on the nomenclature in the config files.

stiber:
> 1. I haven't made the changes in all config files yet.

Making sure that @jbrown725 is looped in, as he is setting up regression testing, which involves config files used for runs within GitHub actions. Come to think of it, do pull request commits trigger actions from the pull request's branch, or from the master branch (if those two differ)? I'm not sure.

> 2. The initialization and seeding assumes that `Norm` assumes that `m = 0.0`, and `s = 1.0`. Is this an okay assumption to make? Should I create rng `loadParameters()` functions that specifies these?

I think that makes sense. Config files should have parameters couched in the domain being simulated, not in terms of lower-level/more abstract things like "raw" means/variances. Let's have that be driven by future needs.

> 3. I need feedback on the nomenclature in the config files.

I think we talked about this, but I'll look at them.

> 4. Should I create an `IRNG` or something to make things more versatile? I thought I'd stick with keeping `MTRand` at the top to reduce the polymorphism performance overhead.

I think your decision is fine for now. If we move to using more than one engine in the future, then this can be revisited.

Simulator/Core/Simulator.cpp
```diff
@@ -83,7 +84,16 @@ void Simulator::loadParameters() {
...
+
+   // Instantiate rng object 
+   string type;
+   ParameterManager::getInstance().getStringByXpath("//RNGConfig/NoiseRNGParams/@class", type);
+   rgNormrnd = RNGFactory::getInstance()->createRNG(type);
+
+   ParameterManager::getInstance().getLongByXpath("//RNGConfig/InitRNGParams/Seed/text()", rngSeed_);
+   ParameterManager::getInstance().getLongByXpath("//RNGConfig/NoiseRNGParams/Seed/text()", seed_);
+   rgNormrnd->seed(seed_);
+   rng.seed(rngSeed_);
```
stiber, line 96: Let's also give `rng` a clearer/less generic name, consistent with naming conventions.

Simulator/Core/Simulator.cpp
```diff
@@ -288,6 +299,8 @@ BGFLOAT *Simulator::getPSummationMap() const { return pSummationMap_; }
 
 long Simulator::getSeed() const { return seed_; }
 
+long Simulator::getRNGSeed() const { return rngSeed_; }
```
stiber, line 302: And we should make these member variables' and accessors' names clearer.

Simulator/Utils/Global.cpp
```diff
@@ -70,10 +70,10 @@ int g_deviceId = 0;
 #endif // USE_GPU
 
 // A random number generator.
-MTRand rng(1);
+MTRand rng;
```
stiber, line 73: Why is this an object, and `rgNormrnd` is a pointer? Should these be global? Perhaps answering that second question can be deferred to a later time by creating an issue for someone else to address.

Simulator/Utils/RNG/Norm.cpp
```diff
@@ -56,31 +56,56 @@
...
+   sigma = s;
+}
+
+void Norm::seed(uint32_t seed) {
+   Norm::seed(0.0, 1.0, seed);
+}
+
+void Norm::seed() {
+   Norm::seed(0.0, 1.0, 0);
+}
```
stiber, line 111: Perhaps these could be made simpler and more consistent with default values for the method parameters?

VivekGandhi54:
That's a good question about the Github actions. Based on the checks that were run on the `regression-testing` branch, it seems like it runs the branch-specific tests.

@jbrown725 This pull request changes all the config files, but the expected output is the same. So (with the old config files) this branch will seg-fault before outputting anything, but it might be good to try and trigger a failed diff check sometime

Simulator/Utils/RNG/Norm.h
```diff
@@ -55,41 +55,51 @@
...
+   virtual void seed(uint32_t seed) override;
+   virtual void seed() override;
+
 private:
-  // Additional state information
+   // Additional state information
 
-  /*! Which of the pair of pseudorandom numbers was last
-    returned. Says whether we should calculate this time  */
-  bool odd;
```
stiber, line 83: Change member variable names to be consistent with our standards.

configfiles/1_tR_1.0--fE_0.90_10000.xml
```diff
@@ -14,9 +14,14 @@
...
-         <value name="value">777</value>
-      </Seed>
+      <RNGConfig name="RNGConfig">
+         <InitRNGParams name="InitRNGParams">
+            <Seed name="Seed">1</Seed>
+         </InitRNGParams>
+         <NoiseRNGParams class="Norm" name="NoiseRNGParams">
+            <Seed name="Seed">777</Seed>
+         </NoiseRNGParams>
+      </RNGConfig>
```
stiber, line 24: Parameter structure looks good.

stiber (changes requested):
A couple overall questions:

1. Have you compared these modifications with the original, to ensure that we get exactly the same sequence of numbers as before the mods?
2. Are you using the same detailed coding patterns for factory, singleton, parameter loading, and object creation as we're using in the other class hierarchies/subsystems that involve factories?

Good job catching misspellings.

Simulator/Utils/Global.cpp
VivekGandhi54: `rgNormrnd` needs to be created dynamically via the factory. If I kept it as an object, I would be bound to either `Norm` or `MTRand`. `rng` is an object because it's slightly faster and I can't think of a reason for it to be a pointer.

Simulator/Core/Simulator.cpp
VivekGandhi54: I might get rid of `getSeed()` and `seed_` , now that it's being seeded right in `loadParameters()`. Unless there's ever a reason to re-seed.

VivekGandhi54:
> Have you compared these modifications with the original, to ensure that we get exactly the same sequence of numbers as before the mods?

I have, and it works fine. Both objects get seeded correctly, and `rgNormrnd` is initialized to the right class.

> Are you using the same detailed coding patterns for factory, singleton, parameter loading, and object creation as we're using in the other class hierarchies/subsystems that involve factories?

Yup, I basically copied over a factory class and only changed things where required.

VivekGandhi54:
> Have you compared these modifications with the original, to ensure that we get exactly the same sequence of numbers as before the mods?

Yup, they both output the same sequences, and get initialized and seeded correctly.

> Are you using the same detailed coding patterns for factory, singleton, parameter loading, and object creation as we're using in the other class hierarchies/subsystems that involve factories?

Yup, I copied over a factory class and just changed the RNG specifics.

Simulator/Utils/RNG/Norm.cpp
VivekGandhi54: Unfortunately, I'm fairly certain that doing that for functions with the same function names would leave you with an infinite recursion.

Simulator/Utils/Global.cpp
stiber: But isn't it also created by the factory?

Simulator/Utils/RNG/Norm.cpp
stiber: I you had `void Norm::seed(BGFLOAT m = 0.0, BGFLOAT s = 1.0, uint32_t seed = 0)`, then you wouldn't need a separate definition of `seed()`. Is there no way to avoid a conflict with `seed(uint32_t seed)`?

Simulator/Utils/Global.cpp
VivekGandhi54: Nope, `rng` (used for initialization) is always an `MTRand` instance. It's initialization seed was hardcoded in earlier. I could create it using the factory if you'd like though.

Simulator/Utils/RNG/Norm.cpp
VivekGandhi54: I was fairly unclear ^, sorry about that. I can explain what I mean a little better when we meet.

`MTRand::seed()` and `MTRand::seed(uint32_t seed)` are both inherited by `Norm`. At least the second definition needs to be overridden to seed `MTRand *rgNormrnd` correctly.

`void Norm::seed(uint32_t seed = 0, BGFLOAT m = 0.0, BGFLOAT s = 1.0)` can't be used to override either of these.
`void Norm::seed(uint32_t seed = 0) override` compiles, but doesn't work as expected when `seed()` is called. It calls `MTRand::seed()`.

I would rather be more explicit with these definitions.
I'll specify the constant values using `#define`s, so that they're all specified in one place.

Simulator/Utils/RNG/Norm.h
```diff
@@ -21,6 +21,10 @@
 #ifndef _NORM_H_
 #define _NORM_H_
 
+#define  DEFAULT_M      0.0
+#define  DEFAULT_S      1.0
+#define  DEFAULT_SEED   0
```
stiber, line 26: Use `const` declarations, rather than `#define`s, as the latter are untyped.

stiber (changes requested):
One small item.

Simulator/Utils/Global.cpp
stiber: Oh, OK, then this is fine. No need to use a factory.

Simulator/Utils/RNG/Norm.cpp
stiber: OK, given that these need to be overridden, this is fine.

VivekGandhi54:
Changes names:
`rng` -> `initRNG`
`rgNormrnd` -> `noiseRNG`
`getSeed()` -> `getNoiseRngSeed()`
`getRNGSeed()` -> `getInitRngSeed()`
`seed_` -> `noiseRngSeed_`
`rngSeed_` -> `initRngSeed_`

stiber:
> Let me know if these are adequate.
> Name changes:
> `rng` -> `initRNG`
> `rgNormrnd` -> `noiseRNG`
> `getSeed()` -> `getNoiseRngSeed()`
> `getRNGSeed()` -> `getInitRngSeed()`
> `seed_` -> `noiseRngSeed_`
> `rngSeed_` -> `initRngSeed_`

These look good.

BTW, make sure to reply to the review conversations above; it will make it easier for me to see that they are resolved (otherwise, I need to hunt through the files to check).

Simulator/Core/Simulator.cpp
```diff
@@ -83,7 +84,16 @@ void Simulator::loadParameters() {
    ParameterManager::getInstance().getIntByXpath("//SimParams/numEpochs/text()", numEpochs_);
    ParameterManager::getInstance().getIntByXpath("//SimConfig/maxFiringRate/text()", maxFiringRate_);
    ParameterManager::getInstance().getIntByXpath("//SimConfig/maxEdgesPerVertex/text()", maxEdgesPerVertex_);
-   ParameterManager::getInstance().getLongByXpath("//Seed/value/text()", seed_);
+
+   // Instantiate rng object 
+   string type;
+   ParameterManager::getInstance().getStringByXpath("//RNGConfig/NoiseRNGParams/@class", type);
+   rgNormrnd = RNGFactory::getInstance()->createRNG(type);
```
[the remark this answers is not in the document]
VivekGandhi54: No idea what `rg` stands for, but I renamed `rgNormrnd` to `noiseRNG`

Simulator/Core/Simulator.cpp
VivekGandhi54: Renamed `rng` to `initRNG`

Simulator/Utils/RNG/Norm.h
VivekGandhi54: Replaced to `static constexpr` declarations.

Simulator/Utils/RNG/Norm.h
VivekGandhi54: Added `_` after all member variables, but didn't change the names further,

Simulator/Core/Simulator.cpp
VivekGandhi54: Renamed for clarity:
`getSeed()` -> `getNoiseRngSeed()`
`getRNGSeed()` -> `getInitRngSeed()`
`seed_` -> `noiseRngSeed_`
`rngSeed_` -> `initRngSeed_`

VivekGandhi54:
Most of the comments were about these name changes, so I marked them as resolved.
Sorry if this pull request has been more painful than necessary. Lots of tiny details

stiber (approved):

PR #232, opened 2021-07-26
Regression testing workflow

### LINKED ISSUE #110: Develop a regression testing workflow
Right now, I'm thinking that we would use test-tiny.xml (or something similar), outputting XML so that we can produce a reference simulation result and then use tools like diff to see if new results match old ones. I think this is feasible for simple simulations on the CPU. Exact reproducibility on the GPU may be more difficult, as it's possible that threads associated with the same neuron in two different simulations may get different RNG-generated noise in two different simulations. But we should be able to take a first step towards this relatively straightforwardly.

jbrown725:
Save known good simulator outputs in a dedicated folder and compare them to outputs generated by our GitHub action workflow. Closes #110


.github/workflows/tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Unit Tests
+name: Regression Tests
```
stiber, line 1: Well, the Google tests are unit tests. Maybe "Unit and regression tests"? Then, you can add "regression" to the names of your three tests below.

.github/workflows/tests.yml
```diff
@@ -21,6 +21,22 @@ jobs:
...
     - name: build
       run: make
-    - name: run tests
-      run: ./tests
+    - name: run unit tests
+      run: ./tests    
+
+    # simulator runs
+    - name: run test-tiny
+      run: ./cgraphitti -c ../configfiles/test-tiny.xml
```
stiber, line 29: I'm a little concerned about leaving the config files for regression testing in the configfiles directory; various people modify them for various purposes (this is not good practice, but...). Perhaps you could copy them into the `../Testing/RegressionTestComparison` directory?

stiber (changes requested):
See comments above.

.github/workflows/tests.yml
jbrown725: My thinking was that unit tests and simulator runs (functional tests) are both under the category of regression testing. Both types are run after every change to make sure nothing breaks.

.github/workflows/tests.yml
jbrown725: test config files have been copied to their own folder in the Testing directory

.github/workflows/tests.yml
stiber: True, but regression testing involves comparison over time, while the unit tests are comparisons with something like "ground truth", such as ensuring that correct files and objects are created.

stiber (approved):
Good to go!

PR #233, opened 2021-07-28
Switch most of the simulator over to valarrays

VivekGandhi54:
Implemented valarrays across the Simulator, where feasible.
I didn't touch the GPU code (other than to edit the references to the member variables) or anything in Utils.

Might be easier to review this commit-by-commit. I realize that the commit messages aren't the most helpful, but I went from directory to directory.


VivekGandhi54:
Reverted changes to `EdgeIndexMap` because CUDA wanted normal C++ arrays

VivekGandhi54:
`test-small.xml` and `test-medium.xml` create the exact same output as `master`. Both run at comparable speeds with `-O3`. Times for `test-medium`:

With valarrays (valarray-switch):
```
real    6m51.058s
user    4m47.936s
sys     1m54.666s
```

With arrays (master):
```
real    6m52.817s
user    4m48.724s
sys     1m55.359s
```

VivekGandhi54:
`test-small.xml` and `test-medium.xml` create the exact same output as `master`. Valarrays seem to be about as fast as arrays, with `-O3`. Times for `test-medium`:

With valarrays (valarray-switch):
```
real    17m12.246s
user    17m10.818s
sys     0m0.497s
```

With arrays (master):
```
real    17m16.532s
user    17m15.025s
sys     0m0.658s
```

VivekGandhi54:
While this works fine, in hindsight, it might be better to go with vectors. I'm not sure if they're any slower, but they seem like a more future-proof choice.

stiber:
I don't think this is something that should be done without a decent knowledge of the GPU code. Additionally, because it impacts the GPU, I think we should avoid doing this until we have verified GPU simulations against matching CPU simulations.

I'm not sure that there should be a difference between valarray and vector, regarding the future of C++. Valarray is much more numerically-focused. I don't think that either class will get deprecated.

VivekGandhi54:
Deferred to later

PR #235, opened 2021-07-29
Fix hdf5

### LINKED ISSUE #234: HDF5Recorder::compileHistories causes an assertion error
HDF5 output works fine on `test-tiny-hdf5.xml`, but when I made a version of `test-small` that uses the `HDF5GrowthRecorder`, it fails an assertion at line 268 of `HDF5Recorder`:
` assert(idx1 >= 0 && idx1 < simulator.getEpochDuration());`

Some notes:
- The problem is that `idx1` is >= `simulator.getEpochDuration()` on a simulation with more than 1 epoch. 
- This assertion is to prevent `burstinessHist_[idx1]++` on the next line from writing outside of bounds, since `burstinessHist_` has a length of `simulator.getEpochDuration()`.
  - Two lines down, `spikesHistory_[idx2]++` is similarly at risk of writing out of bounds, but doesn't have its own assertion.
- Removing this assertion predictably causes a segmentation fault.
- Removing the assertion and arbitrarily making both `burstinessHist_` and `spikesHistory_` twice as long allows `test-small` to complete and write to a .h5 file successfully.

So the solution would be to either initialize these two arrays to a larger, hopefully non-arbitrary value OR figure out why `idx1` and `idx2` are being calculated as values too large for their arrays to hold them. If `idx1` and `idx2` are being calculated correctly, then the array needs to be bigger to fit them.

Also:
- This assertion does not exist in `XMLRecorder`
-  `burtsinessHist_` and `spikesHistory_` are of type `VectorMatrix` in `XMLRecorder`, so dynamic resizing means writing out of bounds is not an issue.

Another solution might be to change these two arrays to both be a `VectorMatrix` in `HDF5Recorder`, but that might not be possible due to the H5 write method wanting an array as an output buffer.

jbrown725:
Change Hdf5Recorder to initialize `burstinessHist_` and `spikesHistory_` as arrays large enough to hold data for multi-epoch simulations. Closes #234.

Incremental HDF5 file writing from BrainGrid was already implemented in d4dc1240a9fb1f8332466846f6012b2a93f1e887. I verified that it was writing readable files using the `h5dump --header` command. I also checked the output file of a simulation in progress and saw that the file size increases after every epoch.

### DIFF
--- a/Simulator/Recorders/Hdf5Recorder.cpp
+++ b/Simulator/Recorders/Hdf5Recorder.cpp
@@ -31,8 +31,8 @@ const H5std_string  nameProbedNeurons("probedNeurons");
 
 /// The constructor and destructor
 Hdf5Recorder::Hdf5Recorder() :
-      offsetSpikesProbedNeurons_(nullptr),
-      spikesProbedNeurons_(nullptr) {
+   offsetSpikesProbedNeurons_(nullptr),
+   spikesProbedNeurons_(nullptr) {
 
    resultFileName_ = Simulator::getInstance().getResultFileName();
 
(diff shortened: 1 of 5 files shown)


jbrown725:
It looks like some commits from my regression testing branch made their way into this PR. Still learning the finer points of git.

.gitignore
```diff
@@ -80,3 +80,5 @@ Output/Debug/*.txt
 # Result files
 Output/Results/*.xml
 Output/Results/*.h5
+Testing/RegressionTesting/TestOutput/*.xml
+Testing/RegressionTesting/TestOutput/*.h5
```
stiber, line 84: Maybe just the whole directories (`Results` and `TestOutput`)?

Simulator/Recorders/Hdf5Recorder.cpp
```diff
@@ -169,20 +171,20 @@ void Hdf5Recorder::initDataSet()
...
 
     // allocate and initialize data memories
-    burstinessHist_ = new int[static_cast<int>(Simulator::getInstance().getEpochDuration())];
-   spikesHistory_ = new int[static_cast<int>(Simulator::getInstance().getEpochDuration() * 100)];
-    memset(burstinessHist_, 0, static_cast<int>(Simulator::getInstance().getEpochDuration() * sizeof(int)));
-    memset(spikesHistory_, 0, static_cast<int>(Simulator::getInstance().getEpochDuration() * 100 * sizeof(int)));
+    burstinessHist_ = new int[static_cast<int>(simulator.getEpochDuration() * simulator.getNumEpochs())];
+    spikesHistory_ = new int[static_cast<int>(simulator.getEpochDuration() * simulator.getNumEpochs() * 100)];
+    memset(burstinessHist_, 0, static_cast<int>(simulator.getEpochDuration() * simulator.getNumEpochs() * sizeof(int)));
+    memset(spikesHistory_, 0, static_cast<int>(simulator.getEpochDuration() * simulator.getNumEpochs() * 100 * sizeof(int)));
```
stiber, line 177: I assume this is the place where the epoch data buffers were reallocated to be much larger. If that isn't the case in BrainGrid, then it must be the case that there is code in BG that avoids the problem we're seeing in Graphitti. Rather than hack this solution, we should find where the missing pieces are in Graphitti and fix them. Otherwise, not only do these become enormous for long simulations, but there will be incompatibilities with the GPU code.

stiber (changes requested):
Hopefully, my comments are coherent. I'm a bit tired this night. 😪

jbrown725:
For Hdf5Recorder:
- Reverted `burstinessHist_` and `spikesHistory_` arrays to original lengths 
- Added back a term in the calculation of idx1 and idx2 that was removed in b82c9d33e3cc3dfba6467b9aff6e5eedce309d25. This term keeps the indices within the bounds of the arrays.

For Hdf5GrowthRecorder:
- Shortened the dimensions of `ratesHistory` and `radiiHistory` dataspaces
    - Their dimensions were made bigger in 0b9f08e1478a58a738c5e10aa625d8e53a45a27c but I'm not sure why. They had a lot of empty space in them

HDF5 files now output without hitting any assertions or seg faults. I verified the outputs by converting them to xml with the `h5dump --xml` command and diffing BrainGrid vs Graphitti outputs. They're all identical now.

Simulator/Recorders/Hdf5Recorder.cpp
stiber: Given fixes in following commits, looks good.

stiber (approved):

PR #278, opened 2021-11-12
Added RecorderFiles layer to config files

kimjamesuw:


king-shak:
Assuming the test passes, all looks good. However, I'd like to test these changes with Workbench first before approving - please don't merge till then.

king-shak (approved):
All looks good! I've tested the changes in Workbench and everything works!

Issue #279, opened 2021-11-15
Simplify or encapsulate spike recording in neuro vertex classes


stiber:
This is a sub-issue of #240 

Right now, extracting spike information for recording in XML or HDF5 is very complex. If we want to simplify the Recorder classes, then we need to have an interface to the spike history information that is straightforward, like a one or two-dimensional array. Morerover, the documentation of the spike history data structures in the vertex classes (inherited from AllSpikingNeurons) is very poor — basically not possible to understand without puzzling through the code (which is what documentation is supposed to obviate). Need to go through AllSpikingNeurons, figure out and document what is going on, and then determine if the current approach can be simplified within AllSpikingNeurons.

If that is not the case, then we should create a wrapper object around the spike history info that presents a simple interface, consistent with either a 2D array, vertex, VectorMatrix, or the like, so that Recorders can access and copy out spike information with code consistent with other variables to record.

Issue #240, opened 2021-08-21
Recorder class re-engineering


stiber:
Right now, the Recorder classes are implemented specially for each model. However, the fundamental operation of a recorder is as follows:

1. A set of variables is specified for recording. Each variable has a name and dimensions (it may be the case that one of the dimensions is unknown ahead of time, for example, for the list of spike times produced by neurons — we know how many neurons, but we don't know how many spikes each neuron will produce).
2. Each variable has a temporal resolution. Some variables need to be recorded (can update) every time step; some will not update more often than once per epoch.
3. In addition, as set of constants may be specified. These are values that do not change during the simulation, such as neuron types, vertex locations, etc.
4. The recorder has an output format, which is captured by the subclass, currently XML or HDF5.

Can we do something like the ParameterManager, where we register variables with a recorder object, providing whatever information is needed at that time (variable name as a string, variable update interval, variable dimensions)? The recorder would have a method that is periodically called to buffer/output variable contents. Using the above scheme, perhaps we would no longer need to provide a list of vertices to attach recorders to, since recording would be at a variable basis and so could be completely generic (we could record edge variables, for instance).

stiber:
Some questions to answer related to this:

- When does the recorder method that periodically captures data to the output get called?
- Who calls it?
- Right now, how does recorder know which variables to capture?
- For capturing data from arrays, how does the recorder know the start and end indices?
- Does the number of items to capture for arrays change during the simulation?

stiber:
Here is the start of this process:

![IMG_0734](https://user-images.githubusercontent.com/1878411/141853039-2a717054-51ab-4d9e-9b38-1bbac52974c6.jpeg)

stiber:
Also includes #141

stiber:
These whiteboard images might be relevant:

![IMG_1246](https://user-images.githubusercontent.com/1878411/198842638-b7ed54a2-82a8-4f09-a9e4-c10f5edb7cee.jpeg)

![IMG_0954](https://user-images.githubusercontent.com/1878411/198842664-37cabd81-fecb-4a4d-905b-40767fe27e36.jpeg)

![Untitled picture](https://user-images.githubusercontent.com/1878411/198842832-b827605c-7562-43e3-8ce5-cc2d8a9df7d8.png)

PoojaPal2021:
[...]

PR #281, opened 2021-11-19
Encapsulate spike recording in neuro vertex classes

stiber:
Pull request for issue #279 

See documentation in docs/Developer/spikeRecording.md in this branch for on-the-fly notes regarding the current state of planning for this redesign. It seems like the approach to be used involves encapsulating the spike recording data structures within a new class that provides a clean interface to the recorders for their need, a clean interface to the CPU side vertex and edge classes, and the ability to allocate and copy the "raw" data structures on the GPU side.


kimjamesuw (approved):
Thanks Professor Stiber for adding all the information in spikeRecording.md!

stiber:
> Thanks Professor Stiber for adding all the information in spikeRecording.md!

No problem. This pull request isn't ready to close yet. I don't think we can disconnect this from resolving #240; I intend this to at least be where we work together on the re-engineering. I'll do the encapsulating and perhaps what you will do is the first step in re-engineering, where you make mods to the recorder classes so that they will work with the changes.

stiber:
Seems like we can do a relatively mild rewrite of the existing Recorder classes' `compileHistories()` methods to use the new `EventBuffer` interface. We will need to keep the `vector` of `EventBuffer`s as a public member of the `AllSpikingNeurons` class temporarily. The we can test and merge this pull request.

jardiamj:
I looked at this to try to understand how the history of events is currently recorded. I understand that the EventBuffer will allow us to look at events that happened in past epochs. Now, we still need the variables in the struct used for the GPU code to be arrays (pointers) since those are potentially updated in GPU code. From that, I assume that when copying the data back from the GPU we would have to insert the new history into the EvenBuffer since the data types are now different.

stiber:
And, in fact, we will always need arrays on the GPU side. Pooja and I had a great conversation on Monday. I think we're heading towards an idea where all of the things that are currently arrays on both the CPU and GPU side will become a subclass (of our own making) of `vector` on the CPU side. Then (and, Pooja, this is a new thought I've had, so it may be crazy or unworkable), the CPU side class could have methods to copy the internal data to/from the GPU (maybe also have methods to allocate/deallocate GPU storage, and maybe a data member to hold the GPU side address within the object). For the basic subclass, this would be very simple — just using the pointer to the internal array returned by `vector::data()`. For the event buffer subclassed from that, it would be more complicated (whatever we need to do, depending on the details of the two implementations).

I'd like to get closer to a situation where major simulation operations are handled by the operation manager object, and it might be the case that the above would allow us to have top-level operations to command copying data to/from the GPU and such copying would happen everywhere via these objects. Not sure about this yet. But we definitely want something like this (copy everything from GPU to CPU) command capability at the very end of the simulation, if we want to be able to guarantee the ability to completely serialize out the state of the simulation. Come to think of it, that would also require commanding a copy of everything from CPU to GPU at the start of a simulation, if we deserialize.

Regardless, baby steps first.

PoojaPal2021:
The changes in the commits from this pull request are accommodated in #334.
Conversation at this pull request might be useful later

Issue #185, opened 2021-04-30
Need to resolve how to accommodate Doxygen within repo


stiber:
We don't want to deal with tracking changes, but we do want doxygen files in the repo.

Might be some clues here: https://gist.github.com/francesco-romano/351a6ae457860c14ee7e907f2b0fc1a5

stiber:
Seems like this was resolved with #286

PR #286, opened 2021-11-25
Created Github Action to auto-generate Doxygen documentation and deploys it to gh-pages branch.

### LINKED ISSUE #253: Regenerate Doxygen for fall 2021
@cherrybiird, please contact me to set up a time to talk about this.

jardiamj:
Implements #185 
Closes #253 

This Github Action auto-generates the project's Doxygen documentation and deploys it to an orphan gh-pages branch.
The idea is that we can set the docs directory in gh-pages branch as the root for our documentation page: https://uwb-biocomputing.github.io/Graphitti/

Does it make sense to have it executing on every push and pull requests to the master branch?

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,12 +1,9 @@
 name: Github Pages
 
 on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+  schedule:
+    # Triggers at the end of the day on the 1st of every month
+    - cron: '0 0 1 * *'
 
 jobs:
   deploy:
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,26 @@
+name: Github Pages
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: mattnotmitt/doxygen-action@v1
+        with:
+          doxyfile-path: docs/Doxygen/Doxyfile
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs
+          destination_dir: docs
+          force_orphan: true


stiber:
I'm not clear about what it would mean to deploy to an orphan branch. My understanding was that GitHub pages wiki only works in master. Regarding how often to do this, we had thought that this would be done quarterly, but if it's automated then it could be done monthly. But I think it would be a waste to do it with every push.

jardiamj:
> 

An orphan branch is a branch that is completely separate from the master (or any other) branch in the repository. Github Pages is a hosting service for static pages. They tend to use a gh-pages branch which is often set as an orphan branch. This is what is being done in the link you posted on issue #185 (https://gist.github.com/francesco-romano/351a6ae457860c14ee7e907f2b0fc1a5) with the only difference that this is not using Travis-ci but a Doxygen Action I found in the Github actions page. The action generates the documentation under docs/html using the doxigen file, and then "deploys/copies" the entirely docs directory into the gh-pages docs directory. This is the only thing in that branch so far but any other html or markdown files can be added to the root directory if we wanted to.
If you go to the repo settings there is now a "pages" section on the left hand side (this seems to be relatively new). There, you can select gh-pages/docs as the source for the project's Github Pages. That is what is being serve on my fork of Graphitti (https://jardiamj.github.io/Graphitti/).
We might not really need the deploying step since, at the moment, it's just the docs directory being served from a different branch. I can remove it if we think that we will not need it.
I will change the action triggering to a scheduled cron job every 1st of the month.

stiber:
OK, I guess I'm going to be a troublemaker. I can see that doxygen will be run monthly; that looks good. What about the rest of our documentation (the markdown pages)? Will we maintain all of that on the gh-pages branch, or will the gh-pages branch get recreated each month from some other branch in which we manually edit the md files? If the former, does that mean that the md files are permanently on the gh-pages branch? Also, in that case, I assume that the copying of the new doxygen stuff completely replaces whatever was there previously. Finally, you mention a "deploying" step; what is that?

jardiamj:
Hahaha! It is the latter. I looked into the action's source code today and it does an actual `git push` into the gh-pages branch, so not a complete copy/overwrite. This is what I mean by deploying (because that's the term used in the GH action).
This workflow is used when writing documentation using a static html generator (Jekyll, Hugo, etc) where you write your documentation in markdown/templates, generate the static html, then "deploy" to a gh-pages branch. We don't really need it since recreating the doxygen documentation in place would suffice. Also, as it stands the gh-pages branch would be updated once a month which might not be ideal for the rest of the documentation.

One reason to have a github pages is that we could use a custom domain but even for this, it might not be necessary. For our purposes, having the Doxygen docs recreated in place once a month is enough. Should I do that?

stiber:
I guess we'll have to talk about this. I don't understand what would be in the separate branch, how we could guarantee that the doxygen documentation doesn't end up being a combination of old and new files, how we could guarantee that the manually written markdown files don't get lost, or what our workflow would be to manually edit our documentation, including situations where we want to edit the documentation in parallel with writing/testing code.

jardiamj:
@stiber, I did a more thorough testing of this action. It checks out the code from our master branch, then generates the doxygen html files from scratch, and finally commits those files along with any changes under our `docs` directory to gh-pages. The commit to gh-pages is set to be orphan so there won't be a commit history in there.

I tested that:
- We can remove the `html` directory under `docs/Doxygen`; since it will be auto-generated we don't need it in our `master` branch.
- Changes to our Graphitti documentation under `docs` will be committed to gh-pages when this action runs. I tested editing, adding, and removing files.

Considering that, if we use do this gh-pages will be our wiki; would we want it running more often than once a month?

stiber:
I don't think that we need to run this more than once a month, and we can always increase the frequency if that turns out to be problematic.

So, we would move all of the existing Markdown files to this branch, too? I guess we can at first copy everything there and run this action manually to double-check that all look good before switching our GH pages officially over to that branch. We can later delete the md files from main.

jardiamj:
The markdown files stay in our main branch, changes to them will be committed to the gh-pages branch whenever this action runs.

stiber (approved):
This is very nice. Let's make it so, Examine the gh-pages branch that it produces, and if all goes well, switch our official pages over. Great job!

PR #294, opened 2021-12-08
Change runtimefiles location

rashwini21:
Changed Runtime files location in 71 files and successfully verified output before and after changes with CPU and GPU.


stiber (changes requested):
- The log files (such as `build/Output/Debug/edges.txt`) need to be excluded from commits to GitHub by editing `.gitignore`.
- Similarly, simulation outputs should be ignored (I believe all contents of the `Results` directory).
- Was doxygen re-run? I'm not asking for that to happen; I'm just wondering about the changes in just a few doxygen files.
- Why are there Markdown files under the build directory? What's in them?

rashwini21:
> * The log files (such as `build/Output/Debug/edges.txt`) need to be excluded from commits to GitHub by editing `.gitignore`.
> * Similarly, simulation outputs should be ignored (I believe all contents of the `Results` directory).
> * Was doxygen re-run? I'm not asking for that to happen; I'm just wondering about the changes in just a few doxygen files.
> * Why are there Markdown files under the build directory? What's in them?


Thank you Professor for this review. The Runtime files location  like /Output/ and /RuntimeFiles was not updated in the .gitignore file. This resulted in not deletion of edges.txt and *.xml output files.  Now even if the .gitignore file is now edited and commited again , it is not able to delete the content *.txt and *.xml files from the output folder. 

There are multiple articles in git, which provide remedy for this situation , like command "git rm --cached FILENAME" .  But these command clear cache and other types of files with it as well. I also experimented with commands like revert. But in order to make add minimum changes to the Master branch. Closing this pull request and continuing this work on another pull request " Changing runtimefile location two #295"

stiber:
I think that the command you indicate would have been fine to untrack those files, but I'll look at the other PR.

Issue #304, opened 2022-01-21
Feature Suggestion: The Usage of `vcpkg`


valenotary:
# What is `vcpkg` ? 
`vcpkg` is a C++ package manager, available on a majority of popular platforms (including Windows, Linux, and Mac).

# What convenience would it bring to Graphitti ? 
Automatic compilation of the correct platform-specific version of a given library, easing burden of cross-platform users of the project. For example, in relation to #302, `log4cplus` does not automatically build for Windows despite the included repo indicating `MSVC` support.

# Does `vcpkg` support each library we currently manually include in Graphitti ?
Unfortunately no :(. 
The packages that _are_ inside `vcpkg`:

- [`cereal`](https://vcpkg.info/port/cereal)
- [`log4cplus`](https://vcpkg.info/port/log4cplus)
- [`gtest`](https://vcpkg.info/port/gtest)  
-[`hdf5`](https://vcpkg.info/port/hdf5)
And the packages that _are not_ inside `vcpkg`:
- `paramcontainer` (this is really old btw)
- `tinyxpath` (this is also really old)

# What would this look like in Graphitti ?
A `vcpkg.json` file would list Graphitti's project dependencies that can be resolved by `vcpkg`. the `vcpkg.cmake` toolchain can be readily included in the `CMAKE_TOOLCHAIN_FILE` variable, or if we are using `Cmake(User)Presets` as in #303, then we can include the toolchain there.
Furthermore, the `ThirdParty` directory could have a slight refactor to better illustrate which dependencies are built by `vcpkg` and which ones are built by us ourselves. An example file directory would look like:
```bash
../Graphitti/
|->...
|-> /ThirdParty/
||-->/vcpkg/ # this is not what the actual install tree looks like, but just short-handing it for simplicity
|||--->log4cplus 
|||--->cereal
|||--->gtest
||-->/manual/
|||--->/paramcontainer/
|||--->/TinyXPath/
|||--->CMakeLists.txt
|->vcpkg.json
|->CMakeLists.txt
|->CMake(User)Presets.json # assuming we implement #303 
|->...
```
# What are some other things to consider ? 
`vcpkg` is another dependency that the user must install themselves (unless we wrap that up in a bash script) - it's not too much trouble to install though. And it can be reused for other projects, not just Graphitti.

One thing to investigate though is how `vcpkg` will affect CI testing in Graphitti - we ideally do not want to include the built `vcpkg` packages into the main git repo.

[...]

Issue #303, opened 2022-01-21
Feature Suggestion: The Usage of `CMake(User)Presets`


valenotary:
# What are `CMake(User)Presets` ?
A feature supported in `CMake 3.19` and onwards that allows projects to save a set of configurations in a `JSON` file to ease the transition between different build types (e.g. Windows/Linux, CPU/GPU, Debug/Release, etc...). Not only can it be called from the command line (via `cmake --preset "<preset-name>"`) and the `cmake-gui`, but `CMake(User)Presets` is readily supported on popular IDEs like Visual Studio (Code). Here is an example snippet to demonstrate:

![image](https://user-images.githubusercontent.com/32428785/150601578-33270446-f762-42b0-94d3-18ff6a21ccf8.png)

On my setup (Windows, Visual Studio 2022), this particular `CMake` project has support for the `CMakeUserPresets.json` file, and detects a subset of configurations for me to choose (those that are supported by my system). If I were to change the environment, such as tapping into my `WSL` distro, the IDE readily changes the subset of available configurations:

![image](https://user-images.githubusercontent.com/32428785/150602024-93c37d37-2f0c-4015-8da6-7c8aec483e38.png)

For some more examples of how these configurations look, please look at the [official documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).

# Main Benefits for Graphitti Developers 

- A project-wide set of configurations that are, for the most part readily set; no more needing to manually changing environment variables inside the `CMakeLists.txt` to change between CPU and GPU modes. Primarily the only variables that may need manual tuning are those that require user-specific paths, such as an installation directory.
- more cleanly specify the default build directory with `binaryDir`
- Cross-xxx support: building a particular configuration of independent factors is simply the composition of those factors. The primary configuration presets we could feasibly see in Graphitti are as follows:
  - `CPU`/`GPU`
  - `Windows`/`Linux`(/`Apple`?)
  - `Debug`/`Release`
  - Generators (`MSVC` versus `Ninja` versus `Makefiles` versus `Xcode`...)
  - Specific Compilers (`cl` versus `gcc` versus `clang`....)
Creating a CPU+Windows+Debug configuration is simply the union of the three individual configurations presets. In other words, we can write 6 individual configurations that handles the specifics of each, and then create configurations that simply inherit from the original 6.

# What would this immediately affect in Graphitti currently ?
- Inside `CMakeLists.txt`:
[...]

Issue #302, opened 2022-01-21
True Cross Platform (Windows) Support


valenotary:
Graphitti should be able to run on any 64-bit device - primarily _Windows_, but I do not see why not on Apple as well. The primary development and testing environments will be Windows and Linux though.

Here are some current points:
- [ ] The usage of  `__invoke` in `Generic FunctionNode.h` is Linux specific; the standard library call is `std::invoke` and is only available in `C++17` _and beyond_. Thus, this would create an explicit need to raise the standard required by the project to `C++17`. This opens up a number of other modernization opportunities, but we should be careful that raising the standard does not produce any _heavy_ breaking changes (afaik, they shouldn't, at worst just a lot of warnings).
- [ ] Immediately, when raising the standard, [there seems to be an issue raised that is prevalent in the Windows SDK that prevents us from compiling](https://developercommunity.visualstudio.com/t/byte-ambiguous-symbol-build-error-after-enabling-s/822269). I am pretty sure this is because the `Timer` class uses stuff from `windows.h` for some reason, but the workaround temporarily, especially if we want to keep using `Timer`, is to get rid of any `using namespace std;` calls around the project; this workaround allows us to bypass this issue in the Windows SDK. _this change also does not introduce any breaking changes into Graphitti on the Linux version_.
- [ ] Current issue is that the `log4cplus.lib` file is not being found on the Windows build, presumably because the `CMakeLists.txt` file may not be building the Windows version automatically (it _should_ be the case that the 3p module _automagically_ builds for the right platform). Building this current version seems to have no issues in compiling _on Linux_. So either some time needs to be spent in looking at the right places for `log4cplus` to be built from source properly per platform (which I currently can't tell if its the issue of the 3p module itself or us as the users), or using a package manager (like `vcpkg`) to take care of that stuff for us.

stiber:
Actually, Mac already works. It seems clang is very close to g++, in that respect.

Note that Graphitti is a 32-bit program, as all GPUs (as far as I know) are 32-bit hardware.

stiber:
We should replace `Timer` with more modern timing calls (this is for performance measurement, I assume). There may be another issue about this; if not, I have example code.

valenotary:
Last time I checked on the CUDA docs, the CUDA toolkit is explicitly 64-bit only nowadays, and as such you can't develop a CUDA application that's not on a 64-bit system. Supposedly you can still target to build a 32-bit application though. 
[...]

PR #306, opened 2022-01-26
Windows, CMakePresets, and vcpkg integration

valenotary:
# PLEASE DO NOT TRY TO ACTUALLY MERGE THIS INTO MASTER YET FOR THE LOVE OF GOD : ) <3

I'm just posting my fork's current work in trying to get Windows supported for Graphitti. 
Some main changes so far, at the time of writing:
- As of #302 , I removed all `using namespace std;` usages as a temporary fix to the old Windows SDK C++17 breaking change - I expect that in a future iteration, we should just attempt to remove using `windows.h`  in `Timer.h` and see if we can move forth with reintroducing `using namespace std;`.
- Moved some constructor assignments into the respective initializer lists instead. 
- changed plain `enum`s to `enum class`es for stronger type checking.
- Changed `USE_GPU` to `__CUDACC__` because _I'm pretty sure_ the latter macro is a way to discern between CUDA/Non-CUDA applications already provided by NVIDIA. 

This build currently doesn't compile on Windows because of the third task point in #302 - for those interested in trying this build out, generate the windows .lib of `log4cplus` with MSVC and link it manually : ). I'm 90% sure none of the changes I introduced so far have affected the Linux build though.

Not implemented in yet is:
- #303 (experimental, will only move forward if there is a general consensus for me to try)
- #304 (same as ^)


valenotary:
Also I should add that the C++ standard has now been raised to 17 in my particular build, and I changed the minimum CMake requirement to `3.21` in preparation for CMakePresets, but since that isnt integrated yet you are free to change the minimum back.

stiber:
Closed due to age.

PR #311, opened 2022-03-17
Created plantUML diagrams and github action to auto-generate the images

### LINKED ISSUE #305: Document top-level program flow
The whiteboard image below is an old summary of the top-level simulator execution sequence. We need to produce an up-to-date version of this, possibly as a UML sequence diagram but maybe not. We also need to incorporate into this information regarding when each kind of class gets instantiated, when each kind of class reads its parameters, and when each kind of class initializes itself.

Top Level Sequence

jardiamj:
Closes #305

The GitHub action is set to run on push and pull requests but the diagrams will only be generated for the files that were modified in the last commit. What this means is, that you would need all the files you want rendered to have been touched in your last commit before you push or open a pull request. I tried looking for a better solution such as using `github.pull_request.base.sha` but that didn't work on my test.

Link to markdown page: [Sequence Diagrams](https://github.com/UWB-Biocomputing/Graphitti/blob/cfd29236c8d6ec60f703ca9ba3e6c431d1114dbf/docs/Developer/sequenceDiagrams.md)


stiber (changes requested):
Looks great; just link in to the existing developer doc MD page.

jardiamj:
The only caveat of having the generated images in the gh-pages branch only is that, in the master branch, some markdown files will have links to non-existing files.

stiber (approved):

PR #316, opened 2022-04-14
Fixing misdirected link for event buffering

stiber:
Unclear why the HTML in the gh-pages branch at https://uwb-biocomputing.github.io/Graphitti/Developer/ isn't updating from master.


stiber:
@jardiamj, I'd like to resolve why the gh-pages and master branches are different. Changes to the MD files in master don't seem to be reflected in the gh-pages branch.

jardiamj:
This is because the GitHub action that does the deployment to gh-pages only runs on the 1st of each month. It seems sensible to have an action that does that whenever a PR has changes to the Markdown files. I will talk to @NewBobCon about this, it should be straightforward.

stiber:
Thank you. For future consumption, it turns out that this was not a misdirected link. There was a bad/old file, but the link is good and will work once this is deployed to gh-pages.

PR #344, opened 2022-05-28
Removed assignment of device pointer to host pointer that caused segmentation fault

### LINKED ISSUE #333: Segmentation fault thrown at the end of a GPU simulation run
Simulations run with the GPU version of Graphitti throw a `segmentation fault` at the very end. The simulation runs, and the results are saved normally. It segfaults after the elapsed time is printed to the console, which might indicate that the segfault is coming from a destructor somewhere.

jardiamj:
Fixes #333

Fixes the segmentation fault at the end of GPU simulations. See conversation in issue #333 for more details.

### DIFF
--- a/Simulator/Vertices/AllVertices.cpp
+++ b/Simulator/Vertices/AllVertices.cpp
@@ -44,19 +44,18 @@ AllVertices::~AllVertices()
 void AllVertices::setupVertices()
 {
    size_ = Simulator::getInstance().getTotalVertices();
-   
 #if defined(USE_GPU)
    // We don't allocate memory for summationMap_ in CPU when building the GPU
    // implementation. This is to avoid misusing it in GPU code.
    summationMap_ = nullptr;
 
-#else // !define(USE_GPU)
+#else
    summationMap_ = new BGFLOAT[size_];
    for (int i = 0; i < size_; ++i) {
       summationMap_[i] = 0;
    }
    Simulator::getInstance().setPSummationMap(summationMap_);
-#endif // define(USE_GPU)
+#endif
 }
 
 ///  Prints out all parameters of the vertices to logging file.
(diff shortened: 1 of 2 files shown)


jardiamj:
`summationMap_` is always allocated in the CPU by `AllVertices::setupVertices()` and it is a valid allocated pointer up until it gets assigned a GPU pointer by this method. I did a global search for `summationMap_` (the AllVertices property member) and checked each one of the places it appears, it is not being used in the GPU simulation except in the line removed by this PR.

The GPU simulation uses instead the `summationMap_` property member of the `AllVerticesDeviceProperties` struct; there is no copying to or from the CPU.

AllVertices::~AllVertices() tries to deallocate `summationMap_` if `size_` is not 0. Since it is also set by `AllVertices::setupVertices()`, it tries to deallocate the pointer that now points to a GPU memory location.

```
AllVertices::~AllVertices()
{
   if (size_ != 0) {
      delete[] summationMap_;
   }

   summationMap_ = nullptr;

   size_ = 0;
}
```

If we are sure that the GPU implementation doesn't need `summationMap_`, I agree that it would be a good idea to set it to `nullptr` and change the destructor to check that is not `nullptr` instead of `size_ != 0`. But I don't know enough about the purpose of `summationMap_` to make an educated decision.

stiber:
> If we are sure that the GPU implementation doesn't need summationMap_, I agree that it would be a good idea to set it to nullptr and change the destructor to check that is not nullptr instead of size_ != 0. But I don't know enough about the purpose of summationMap_ to make an educated decision.

For the record, I was thinking that we might even use conditional compilation directives so that, for a GPU build, this class wouldn't even have a CPU-side summation map. However, this might have implications for future, full serialization/deserialization, if a class has different member depending on the build type. So, best to just ensure that it is nulled out for GPU simulations and that no storage is allocated for it on the CPU.

@rashwini21, I am CCing you here since you are thinking about the intersection of different CUDA capabilities, software engineering/design considerations, and performance. I don't think there is a performance issue here, but clearly there are issues regarding data types and the design of complex GPU systems with the ability to swap functionality between CPU and GPU builds. I don't know that anyone has addressed this.

stiber (approved):
Looks good!

PR #347, opened 2022-05-30
Created automated format checking script

NewBobCon:
I created a GitHub action to automatically run clang-format during push and pull requests whenever a .cpp or .h file is included. This action has been tested on a separate fork of Graphitti so it should work straight away.


stiber (approved):
Looks good to me. Still hard to believe our code is now formatted consistently!

Issue #331, opened 2022-05-06
Create UML class diagrams using PlantUML


stiber:
Right now, we have hand-drawn class diagrams, which are included as images. They were drawn in OmniGraffle on a Mac, so are not easily modifiable by anyone else and, of course, GitHub can't track changes in the source document. Let's switch to PlanUML, as with other UML diagrams, to eliminate these problems and to generate up-to-date diagrams.

stiber:
Looking at commit #348 we now have PlantUML class diagrams, but they are very flat and not human-friendly. It would be good to have PlantUML diagrams that are more like the hand-drawn block diagram at https://uwb-biocomputing.github.io/Graphitti/Developer/ClassDiagrams/hand-drawn.pdf

PR #348, opened 2022-05-30
Generated UML class diagrams

jardiamj:
Generated UML Class Diagram from current implementation as per issue #331. These diagrams were generated using hpp2plantuml and then grouped into the various components used in Graphitti. There is one large diagram with an overall view of Graphitti, for which I also created a domain diagram for simplicity.

I only included the Neuro classes since NG911 is still a work in progress and the Domain diagram could be improved by grouping the entities into related packages.


NewBobCon (approved):
Looks good to me. Great work!

stiber (approved):

PR #352, opened 2022-06-10
Factory cleanup

### LINKED ISSUE #361: Clean up code of factory classes
All factory classes have a map with key names associated with a pointer to the create method of a given class. Currently, the create method loops through the map data structure to find the correct class; this is unnecessary and unexpected since you can access the class name from a map using `find`.

These classes should be changed to use `find`, if nothing else it would make the code easier to read.

jardiamj:
Closes #361.

I did some cleaned up of the various factory classes. If nothing else, this should improve readability.

Summary:

- Switched to using `find`, instead of looping over the `createFunctions` map.
- Changed `noiseRNG` from a naked pointer to a `shared_ptr`.
- Removed not needed class member pointer to the instance created by the factory.
- Wrote Unit Tests for all factory classes.
- Fixed Segmentation fault on destruction of `Connections911` class (revealed by the unit tests).

### DIFF
--- a/Simulator/Utils/Global.h
+++ b/Simulator/Utils/Global.h
@@ -53,6 +53,7 @@
 extern int g_debug_mask;
 
    #include <cassert>
+   #include <memory>
    #include <sstream>
    #ifdef _WIN32   //needs to be before #include "bgtypes.h" or the #define BGFLOAT will cause problems
       #include <windows.h>                 //warning! windows.h also defines BGFLOAT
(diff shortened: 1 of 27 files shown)


Simulator/Utils/Global.h
```diff
@@ -65,6 +64,7 @@ typedef unsigned long long int uint64_t;   //included in inttypes.h, which is no
    //#include "Norm.h"
    #include "Coordinate.h"
    #include "VectorMatrix.h"
+   #include "memory"
```
stiber, line 67: Should this be `#include <memory>`?

jardiamj: Fixed. It is odd that both include forms work and point to the same file.

jardiamj:
Can we merge this?

stiber (approved):

stiber:
Sorry for the slow response time...

Issue #369, opened 2022-10-18
Modify 911 model to load initial graph structure from a GraphML file


jardiamj:
The previous proof of concept used a grid layout for setting up the initial 911 graph structure. Since the communication links between 911 entities (Caller Regions, PSAPs, and Responders) depend on their geographical location and boundaries, we are now extracting these relationships from a set of Geographic Information Systems (GIS) files into a graph. This graph is saved to a GraphML file that can later be loaded into the Simulator.

We need to modify the corresponding `Layout`, `Connections`, `AllVertices`, and `AllEdges` classes to use the GraphML file to create vertices, edges, index maps, and their properties.

PR #379, opened 2022-11-12
Added ability to read GraphML files into Graphitti

jardiamj:
This PR addresses most of what is needed for reading the initial graph from graphML files. I created a GraphManager class that wraps around some Boost Graph Library functionality.

It addresses most of what is needed for issue #369. Not all the parameters needed for a fully functional 911 implementation are there yet, but that will be addressed in a different PR so as to not make this one too big.


Simulator/Layouts/Neuro/FixedLayout.cpp
```diff
@@ -12,14 +12,39 @@
...
+         (*dist2_)(n, n2) = ((*xloc_)[n] - (*xloc_)[n2]) * ((*xloc_)[n] - (*xloc_)[n2])
+                            + ((*yloc_)[n] - (*yloc_)[n2]) * ((*yloc_)[n] - (*yloc_)[n2]);
+
+         // both points are equidistant from each other
+         (*dist2_)(n2, n) = (*dist2_)(n, n2);
+      }
+   }
+
+   // take the square root to get actual distance (Pythagoras was right!)
+   (*dist_) = sqrt((*dist2_));
```
stiber, line 45: So, why does the superclass own the `dist_` and `dist2_` variables, but they're initialized in the subclass?

stiber (changes requested):
One question to address. Might not require a change.

Simulator/Layouts/Neuro/FixedLayout.cpp
jardiamj: This is a good question. Both `dist_` and `dist2_` are calculated from `xloc_` and `yloc_`, and since their initialization is different in the 911 system, I had to move the initialization of all those arrays to the subclasses. I realize now that this probably breaks `DynamicLayout`, which I am not sure what it does exactly but looking at its code it could inherit from `FixedLayout` and just override `printParameters()`, `generateVertexTypeMap()`, and `initStarterMap()`.

And the question is how to handle class members that are common to all models but whose initialization differs. Should they be part of the base class, or moved to their respective derived classes?

stiber (approved):
If you saw a question in your email, it's because I forgot that `delta_` became a local variable.

Issue #335, opened 2022-05-10
Add variable names to serialization files


stiber:
See the NVP (name-value pair) functions described in the [Cereal documentation](http://uscilab.github.io/cereal/quickstart.html).

PR #382, opened 2022-12-02
[Issue-335] Add names to serialized variables

KamathScience:
This code change focuses on adding variable names to the serialized XML files to improve readability. Variable names are added using Cereal name-value-pair functionality as mentioned in issue #335. 

The code change also improves the size validation checks of deserialized data by adding size variables for each vector in the serialized files. Finally, I have refactored the main method to enhance the testability of Serialization and Deserialization.

### Testing
Since each serialized and deserialized unit test requires a Simulator object (which is a singleton in nature), I had to create a separate main function for each test to avoid sharing the simulator object within the tests- which would otherwise lead to segmentation fault error.


Simulator/Core/Core.cpp
```diff
@@ -0,0 +1,58 @@
+/**
```
stiber, line 1: Let's change this file name to Graphitti_Main.cpp

Simulator/Core/Driver.cpp
```diff
@@ -5,7 +5,7 @@
  *
```
stiber, line 5: Let's change this class to `Core`.

Simulator/Core/Driver.cpp
```diff
@@ -56,38 +49,88 @@
...
-// forward declarations
-bool parseCommandLine(int argc, char *argv[]);
-bool deserializeSynapses();
-void serializeSynapses();
+///  Handles parsing of the command line
+///
+///  @param  argc      argument count.
+///  @param  argv      arguments.
+///  @returns    true if successful, false otherwise.
+bool parseCommandLine(string cmdLineArguments)
```
stiber, line 57: Let's make this a private method.

Simulator/Core/Driver.cpp
```diff
@@ -56,38 +49,88 @@
...
 ///  from parameter file. All initial loading before running simulator in Network
-///  is here.
+///  is done here.
 ///
 ///  @param  argc    argument count.
 ///  @param  argv    arguments.
-///  @return -1 if error, else if success.
-int main(int argc, char *argv[])
+///  @return -1 if error, else 0 if success.
+int Driver::setupSimulation(string cmdLineArguments)
```
stiber, line 120: Let's call this `runSimulation()`.

stiber (changes requested):

Simulator/Core/Core.cpp
```diff
@@ -0,0 +1,58 @@
...
+   // storing command line arguments as string
+   // required to pass as an argument to setupSimulation
+   string cmdLineArguments;
+   for (int i = 1; i < argc; i++) {
+      cmdLineArguments = cmdLineArguments + argv[i] + " ";
+   }
+
+   // Creating an instance of driver class
+   Driver driver;
+   return driver.setupSimulation(cmdLineArguments);
```
stiber, line 57: pass also `argv[0]` as `executableName`.

stiber (changes requested):

Simulator/Core/Driver.cpp
KamathScience: Since the `parseCommandLine ` method need not be referenced outside the `Core.cpp` file (formerly known as `Driver.cpp`), should we give it an internal linkage by declaring it static? - then we don't have to add the private declaration in the header file?

Simulator/Core/Core.cpp
KamathScience: Modified the filename from` Core.cpp` to `Graphitti_main.cpp`.

Simulator/Core/Driver.cpp
KamathScience: Modified the filename `Driver.cpp` and `Driver.h` to `Core.cpp` and `Core.h` respectively.

Simulator/Core/Driver.cpp
KamathScience: Modified the method name from `setupSimulation `to `runSimulation`

Simulator/Core/Core.cpp
KamathScience: Added variable `executableName`

Simulator/Core/Driver.cpp
stiber: Well, this is an interesting question. The use of `static` you indicate is the C usage (and also for ordinary functions). If we make it a private method, then `static` means something different: that it can be called without making an instance of the class. Since we just want to make it executable only by other `Core` methods, that's not necessary.

stiber (approved):
I "unresolved" a conversation, but I believe that that was just a reply to your query and doesn't require any code changes.

Simulator/Core/Driver.cpp
KamathScience: Thank you for clarifying that Professor

PR #384, opened 2022-12-06
 PP :: master build error fix

PoojaPal2021:
This PR contains the fix for the build error in GPU simulation that had caused because of the an old PR #379


stiber (approved):
Good job!

PR #398, opened 2023-01-31
Basic CircularBuffer class

jardiamj:
This is a basic CircularBuffer class intended to be used in the implementation of the NG-911 Discrete Event Queueing model.

In the current implementation, the `get()` method returns a default constructed object when called on an empty buffer. If we decide that it's worth it, we could move our target standard to C++17 and take advantage of the `std::optional` data type. The main advantage is that we will get an obviously invalid value in such a case.


Simulator/Utils/CircularBuffer.h
```diff
@@ -0,0 +1,147 @@
...
+ * the NG-911 system. The capacity of the buffer can be defined on instantiation
+ * and you are only allowed to resize an empty buffer.
+ * 
+ * Trying to put an element into a buffer that is full results in an error,
+ * captured via an assertion. Similarly, if the buffer is empty the value returned
+ * by get() is, naturally, invalid. The onus is on the programmer to check if the
+ * buffer is empty or full before retrieving or inserting values.
+ * 
+ */
+
```
stiber, line 20: `#pragma once`?

stiber (changes requested):
Since you raise the issue of C++17, consult with @KamathScience on her future plans for presenting language features and when the feels that she would be ready to facilitate a discussion of which C++ version we should target. My guess is that it will be C++17, but let's make sure we have heard about the key features at each step and any gotcha's.

KamathScience:
> Since you raise the issue of C++17, consult with @KamathScience on her future plans for presenting language features and when the feels that she would be ready to facilitate a discussion of which C++ version we should target. My guess is that it will be C++17, but let's make sure we have heard about the key features at each step and any gotcha's.

I support upgrading the codebase to C++17 (many C++11 features are refined and polished in C++14 and C++17). The only reservation I have is with GPU - I am not sure if upgrading to C++17 would affect the GPU code/working.

stiber (approved):

PR #402, opened 2023-02-11
Update .gitignore to exclude unit test executable

### LINKED ISSUE #401: Generated unit test executables are not included in .gitignore
The generated unit test executables for serialization and deserialization are not included in `.gitignore` so `git` identifies them as untracked files.

KamathScience:
Modified .gitignore file to ignore the generated unit test executables for serialization and deserialization and its output.
Fixes #401

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,8 @@ graphitti
 ggraphitti
 cgraphitti
 tests
+serialTest
+deserialTest
 *.exe
 *.out
 *.app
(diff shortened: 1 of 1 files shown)


stiber (approved):
Note that you may need to explicitly tell git to stop tracking those file(s). I think that the `.gitignore` just prevents git from tracking changes and/or newly created files.

jardiamj:
The files are currently not being tracked by `git` so adding them to `.gitignore` should be enough.

PR #403, opened 2023-02-13
Switch to Smart-Pointer & STL Containers 

### LINKED ISSUE #23: Switch to smart pointers wherever it makes sense
This will hopefully force us to decide what objects own what data and are responsible for managing it, and which are merely accessing stuff. It will also enable easier serialization using cereal, for the future.

Connects to #24 and #129 and may not be necessary once that is done (all that will be left is use of pointers to support late binding).

### LINKED ISSUE #76: Bring code up to C++11 standard
See issue UWB-Biocomputing/BrainGrid#232 to see the bug aspect of this. Currently, the Makefile is set to compile code using the C++98 standard. There is a problem with initialization of static const member variables, however. Also, generally, there can be some code cleanup and modernization (that might also have a small positive impact on performance). This is a worthwhile project.

## *What* is affected by this?

Simulator

### LINKED ISSUE #129: Change Synapse class arrays to vectors

### LINKED ISSUE #372: Simulator class contains pointers that are never used nor initialized.
The Simulator class contains the following pointers that are never used nor initialized:
- `vertexType *rgNeuronTypeMap_`, and
- `bool *rgEndogenouslyActiveNeuronMap_`

There are getters for these but they are not called anywhere. They should be removed if not needed.

Besides, the class has a pointer (`pSummationMap_`) to the vertices `summationMap_`. The getter for this pointer is called in two places:
- HostSInputRegular.cpp, and
- SInputPoisson.cpp

If we really need to get the vertices `summationMap_` from the Simulator class, we could get it from `model_->getLayout()->getVertices()->summationMap_` rather than having the same pointer in two places. It also presents the assumption that all models have a summation map.

KamathScience:
Fixes #23 
Fixes #372
Fixes #129 

This PR addresses multiple issues as they are mostly interrelated. Some of the changes addressed in the PR are as follows :
1. C++ version has been updated to C++17 in the CMake file
2. Dynamically allocated arrays are changed to vectors
3. Raw pointers and shared_ptr are changed to unique_ptr
4. Use =default notation to mark the constructor and destructor to avoid using an empty body

### DIFF
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -190,11 +190,11 @@ bool Connections911::erasePSAP(AllVertices &vertices, Layout &layout)
       int srcVertex = callersToReroute[i];
 
       int closestPSAP = psaps[0];
-      BGFLOAT smallestDist = (layout.dist_)(srcVertex, closestPSAP);
+      BGFLOAT smallestDist = layout.dist_(srcVertex, closestPSAP);
 
       // Find closest PSAP
       for (int i = 0; i < psaps.size(); i++) {
-         BGFLOAT dist = (layout.dist_)(srcVertex, psaps[i]);
+         BGFLOAT dist = layout.dist_(srcVertex, psaps[i]);
          if (dist < smallestDist) {
             smallestDist = dist;
             closestPSAP = psaps[i];
@@ -220,11 +220,11 @@ bool Connections911::erasePSAP(AllVertices &vertices, Layout &layout)
       int destVertex = respsToReroute[i];
 
       int closestPSAP = psaps[0];
-      BGFLOAT smallestDist = (layout.dist_)(closestPSAP, destVertex);
+      BGFLOAT smallestDist = layout.dist_(closestPSAP, destVertex);
 
       // Find closest PSAP
       for (int i = 0; i < psaps.size(); i++) {
-         BGFLOAT dist = (layout.dist_)(psaps[i], destVertex);
+         BGFLOAT dist = layout.dist_(psaps[i], destVertex);
          if (dist < smallestDist) {
             smallestDist = dist;
             closestPSAP = psaps[i];
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -90,7 +90,7 @@ bool Connections911::updateConnections(AllVertices &vertices)
    int numVertices = Simulator::getInstance().getTotalVertices();
    Layout &layout = *Simulator::getInstance().getModel()->getLayout();
    oldTypeMap_ = new vertexType[numVertices];
-   memcpy(oldTypeMap_, layout.vertexTypeMap_, numVertices * sizeof(vertexType));
+   memcpy(oldTypeMap_, layout.vertexTypeMap_.data(), numVertices * sizeof(vertexType));
 
    // Erase PSAPs
    for (int i = 0; i < psapsToErase_; i++) {
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -29,8 +29,8 @@ void Connections911::setup()
    LOG4CPLUS_INFO(fileLogger_, "Initializing connections");
 
    // we can obtain the Layout, which holds the vertices, from the Model
-   shared_ptr<Layout> layout = Simulator::getInstance().getModel()->getLayout();
-   shared_ptr<AllVertices> vertices = layout->getVertices();
+   Layout &layout = *Simulator::getInstance().getModel()->getLayout();
+   AllVertices &vertices = *layout.getVertices();
 
    // Get list of edges sorted by target in ascending order from GraphManager
    GraphManager &gm = GraphManager::getInstance();
@@ -40,10 +40,10 @@ void Connections911::setup()
    for (auto it = sorted_edge_list.begin(); it != sorted_edge_list.end(); ++it) {
       size_t srcV = gm.source(*it);
       size_t destV = gm.target(*it);
-      edgeType type = layout->edgType(srcV, destV);
-      BGFLOAT *sumPoint = &vertices->summationMap_[destV];
+      edgeType type = layout.edgType(srcV, destV);
+      BGFLOAT *sumPoint = &vertices.summationMap_[destV];
 
-      BGFLOAT dist = layout->dist_(srcV, destV);
+      BGFLOAT dist = layout.dist_(srcV, destV);
       LOG4CPLUS_DEBUG(edgeLogger_, "Source: " << srcV << " Dest: " << destV << " Dist: " << dist);
 
       BGSIZE iEdg;
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -13,14 +13,6 @@
 #include "Layout911.h"
 #include "ParameterManager.h"
 
-Connections911::Connections911()
-{
-}
-
-Connections911::~Connections911()
-{
-}
-
 void Connections911::setup()
 {
    int added = 0;
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -43,7 +43,7 @@ void Connections911::setup()
       edgeType type = layout->edgType(srcV, destV);
       BGFLOAT *sumPoint = &vertices->summationMap_[destV];
 
-      BGFLOAT dist = (layout->dist_)(srcV, destV);
+      BGFLOAT dist = layout->dist_(srcV, destV);
       LOG4CPLUS_DEBUG(edgeLogger_, "Source: " << srcV << " Dest: " << destV << " Dist: " << dist);
 
       BGSIZE iEdg;
--- a/Simulator/Connections/NG911/Connections911.cpp
+++ b/Simulator/Connections/NG911/Connections911.cpp
@@ -43,7 +43,7 @@ void Connections911::setup()
       edgeType type = layout->edgType(srcV, destV);
       BGFLOAT *sumPoint = &vertices->summationMap_[destV];
 
-      BGFLOAT dist = (*layout->dist_)(srcV, destV);
+      BGFLOAT dist = (layout->dist_)(srcV, destV);
       LOG4CPLUS_DEBUG(edgeLogger_, "Source: " << srcV << " Dest: " << destV << " Dist: " << dist);
 
       BGSIZE iEdg;
@@ -190,11 +190,11 @@ bool Connections911::erasePSAP(AllVertices &vertices, Layout &layout)
       int srcVertex = callersToReroute[i];
 
       int closestPSAP = psaps[0];
-      BGFLOAT smallestDist = (*layout.dist_)(srcVertex, closestPSAP);
+      BGFLOAT smallestDist = (layout.dist_)(srcVertex, closestPSAP);
 
       // Find closest PSAP
       for (int i = 0; i < psaps.size(); i++) {
-         BGFLOAT dist = (*layout.dist_)(srcVertex, psaps[i]);
+         BGFLOAT dist = (layout.dist_)(srcVertex, psaps[i]);
          if (dist < smallestDist) {
             smallestDist = dist;
             closestPSAP = psaps[i];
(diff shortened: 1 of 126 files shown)


Simulator/Edges/AllEdges.cpp
```diff
@@ -109,22 +81,16 @@ void AllEdges::setupEdges(const int numVertices, const int maxEdges)
...
-      type_ = new edgeType[maxTotalEdges];
-      inUse_ = new bool[maxTotalEdges];
-      edgeCounts_ = new BGSIZE[numVertices];
+      destVertexIndex_.resize(maxTotalEdges);
+      W_.resize(maxTotalEdges, 0);
+      summationPoint_.resize(maxTotalEdges);
+      sourceVertexIndex_.resize(maxTotalEdges);
+      type_.resize(maxTotalEdges);
+      inUse_ = make_unique<bool[]>(maxTotalEdges);
+      edgeCounts_.resize(numVertices, 0);
```
stiber, line 90: Why are some initialized and some not?

Simulator/Edges/AllEdges.h
```diff
@@ -198,26 +198,26 @@ class AllEdges {
...
+   vector<edgeType> type_;
 
    ///  The boolean value indicating the entry in the array is in use.
-   bool *inUse_;
+   unique_ptr<bool[]> inUse_;
 
    ///  The number of (incoming) edges for each vertex.
    ///  Note: Likely under a different name in GpuSim_struct, see edge_count. -Aaron
-   BGSIZE *edgeCounts_;
+   vector<BGSIZE> edgeCounts_;
```
stiber, line 220: I'm a little concerned about initialization. I _think_ that `resize()` will just initialize the newly added values in a vector, so that would mean that the initially created entries in the default sized vectors might not be initialized correctly.

Simulator/Edges/AllEdges.h
```diff
@@ -298,8 +298,8 @@ template <class Archive> void AllEdges::load(Archive &archive, std::uint32_t con
 
    // assigns serialized data to objects
    for (int i = 0; i < requiredSize; i++) {
-      W_[i] = WVector[i];
-      sourceVertexIndex_[i] = sourceVertexLayoutIndexVector[i];
-      destVertexIndex_[i] = destVertexLayoutIndexVector[i];
+      W_.push_back(WVector[i]);
+      sourceVertexIndex_.push_back(sourceVertexLayoutIndexVector[i]);
+      destVertexIndex_.push_back(destVertexLayoutIndexVector[i]);
```
stiber, line 303: I think that serialization and deserialization can work directly with the vector data members, no longer needing to marshall/un-marshall to/from raw arrays.

Simulator/Layouts/Layout.cpp
```diff
@@ -94,10 +78,10 @@ void Layout::registerGraphProperties()
...
 {
    // Allocate memory
-   xloc_ = new VectorMatrix(MATRIX_TYPE, MATRIX_INIT, 1, numVertices_);
-   yloc_ = new VectorMatrix(MATRIX_TYPE, MATRIX_INIT, 1, numVertices_);
-   dist2_ = new CompleteMatrix(MATRIX_TYPE, MATRIX_INIT, numVertices_, numVertices_);
-   dist_ = new CompleteMatrix(MATRIX_TYPE, MATRIX_INIT, numVertices_, numVertices_);
+   xloc_ = make_unique<VectorMatrix>(MATRIX_TYPE, MATRIX_INIT, 1, numVertices_);
+   yloc_ = make_unique<VectorMatrix>(MATRIX_TYPE, MATRIX_INIT, 1, numVertices_);
+   dist2_ = make_unique<CompleteMatrix>(MATRIX_TYPE, MATRIX_INIT, numVertices_, numVertices_);
+   dist_ = make_unique<CompleteMatrix>(MATRIX_TYPE, MATRIX_INIT, numVertices_, numVertices_);
```
stiber, line 84: We should probably talk about whether the data members should be pointers to these objects or just these objects themselves. I don't remember why they were pointers in the first place...

Simulator/Edges/AllEdges.h
KamathScience: Agreed. Will simplify this.

Simulator/Layouts/Layout.cpp
jardiamj: One thing that comes to mind is that we don't allocate memory for these data members when the Layout is instantiated because at that point we don't yet know the number of vertices for the simulation. That might be the reason these are pointers and not objects.
By the way, the delayed construction of class member objects is another potential use for `std::optional`.

stiber: That's a possibility, and at some level the question comes down to what approach best communicates the purpose of the code ("what going on?") to a human reader, since the various approaches all "work". I don't have the answer right now, but I'll propose a third possibility: declare the data members with default sizes (like 1x1 — or modify the classes to add defaults to the constructors), and then, when we know how big they should be, just use assignment to change them. This would be like using an `int` data member, for example — it might be initialized to zero or whatever in the constructor, but in some other method set to a "better" value.

Simulator/Edges/AllEdges.cpp
KamathScience: All member variables are now initialized.

Simulator/Edges/AllEdges.h
KamathScience: Used assign to assign default values.

Simulator/Connections/NG911/Connections911.cpp
```diff
@@ -43,7 +43,7 @@ void Connections911::setup()
       edgeType type = layout->edgType(srcV, destV);
       BGFLOAT *sumPoint = &vertices->summationMap_[destV];
 
-      BGFLOAT dist = (*layout->dist_)(srcV, destV);
+      BGFLOAT dist = (layout->dist_)(srcV, destV);
```
jardiamj, line 46: I believe the parentheses around `(layout->dist_)` are no longer necessary.

KamathScience: Will fix this, thanks!

stiber (changes requested):
Some items.

Simulator/Connections/NG911/Connections911.cpp
```diff
@@ -90,7 +90,7 @@ bool Connections911::updateConnections(AllVertices &vertices)
    int numVertices = Simulator::getInstance().getTotalVertices();
    Layout &layout = *Simulator::getInstance().getModel()->getLayout();
    oldTypeMap_ = new vertexType[numVertices];
-   memcpy(oldTypeMap_, layout.vertexTypeMap_, numVertices * sizeof(vertexType));
+   memcpy(oldTypeMap_, layout.vertexTypeMap_.data(), numVertices * sizeof(vertexType));
```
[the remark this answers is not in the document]
KamathScience: Agreed, changed oldTypeMap to be a vector.

Simulator/Connections/Neuro/ConnStatic.h
```diff
@@ -67,19 +67,19 @@ class ConnStatic : public Connections {
    }
 
    /// Get array of vertex weights
-   BGFLOAT *getWCurrentEpoch() const
+   vector<BGFLOAT> getWCurrentEpoch() const
    {
       return WCurrentEpoch_;
    }
```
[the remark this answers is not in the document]
KamathScience: Missed this, yes, the best practice is to return a const reference to a vector- just like you pointed out.

Simulator/Connections/Neuro/ConnStatic.h
```diff
@@ -67,19 +67,19 @@ class ConnStatic : public Connections {
...
    {
       return sourceVertexIndexCurrentEpoch_;
    }
 
    /// Get all edge destination vertex indices
-   int *getDestVertexIndexCurrentEpoch() const
+   vector<int> getDestVertexIndexCurrentEpoch() const
    {
       return destVertexIndexCurrentEpoch_;
    }
```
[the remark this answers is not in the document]
KamathScience: Modified this

Simulator/Layouts/Layout.cpp
```diff
@@ -52,30 +45,11 @@ Layout::Layout() : numEndogenouslyActiveNeurons_(0)
 /// Destructor
 Layout::~Layout()
 {
```
stiber, line 47: Do we leave destructors empty, or do we delete their implementations to rely on the default behavior (which is the same thing...)?

Simulator/Layouts/Layout.h
```diff
@@ -67,28 +67,28 @@ class Layout {
...
+   CompleteMatrix dist_;   ///< The true inter-neuron distance.
 
    vector<int>
       probedNeuronList_;   ///< Probed neurons list. // ToDo: Move this to Hdf5 recorder once its implemented in project -chris
 
-   vertexType *vertexTypeMap_;   ///< The vertex type map (INH, EXC).
+   vector<vertexType> vertexTypeMap_;   ///< The vertex type map (INH, EXC).
 
-   bool *starterMap_;   ///< The starter existence map (T/F).
+   vector<bool> starterMap_;   ///< The starter existence map (T/F).
```
stiber, line 83: Will this cause problems with GPU implementation, or is just used during initialization (implemented using CPU code)?

Simulator/Utils/Global.h
stiber, line 1: Lines 12 and 13 (and maybe 199? not sure where the matching `#endif` is) have `#ifdef` style guards against multiple includes, but we already have `#pragma once`.

Simulator/Core/Model.cpp
```diff
@@ -177,23 +178,22 @@ void Model::updateHistory()
 /// Get the Connections class object.
 /// @return Pointer to the Connections class object.
 // ToDo: make smart ptr
-shared_ptr<Connections> Model::getConnections() const
+Connections *Model::getConnections() const
```
stiber, line 181: I'm just putting this in a second place; there are more locations where we have "get" methods that return pointers to objects that we don't implement as formal singletons but which really are (we only create one and use a `unique_ptr` to point to it). Should we return a reference, rather than a pointer? I'm not sure...

stiber (changes requested):
OK, some more thoughts. I guess you have to expect fussing about details when more than 100 files change. Some of this is not even your stuff, but rather stuff that should have been questioned before.

Simulator/Connections/NG911/Connections911.cpp
```diff
@@ -29,8 +27,8 @@ void Connections911::setup()
    LOG4CPLUS_INFO(fileLogger_, "Initializing connections");
 
    // we can obtain the Layout, which holds the vertices, from the Model
-   shared_ptr<Layout> layout = Simulator::getInstance().getModel()->getLayout();
-   shared_ptr<AllVertices> vertices = layout->getVertices();
+   Layout &layout = *Simulator::getInstance().getModel()->getLayout();
+   AllVertices &vertices = *layout.getVertices();
```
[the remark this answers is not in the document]
KamathScience: I agree, passing a reference improves the readability; it is clearer to the reader that the memory is allocated, and a pointer on the other hand could be null. 

Given the scope of this change and how large this PR is getting, I would like to make it a new PR linked to a new issue.

Simulator/Layouts/Layout.cpp
KamathScience: I was in a dilemma about this too. I will go ahead and remove the empty destructor implementation as it is cleaner and mark the destructor declaration =default to maintain readability.

Simulator/Layouts/Layout.h
KamathScience: It did cross my mind, so I did check if starterMap_ was being used in GPU code and it wasn't. Just to be on the safe side do you prefer this to be a unique_ptr to bool[ ] ?

Simulator/Core/Model.cpp
KamathScience: I am more inclined toward returning a reference. Will make that change in the new PR.

KamathScience: From Google Style Guide: "Prefer to return by value or, failing that, return by reference. Avoid returning a pointer unless it can be null."

Simulator/Utils/Global.h
KamathScience: Will check on this

Simulator/Connections/NG911/Connections911.cpp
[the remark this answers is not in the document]
stiber: OK

Simulator/Layouts/Layout.h
stiber: No, this is fine.

Simulator/Core/Model.cpp
stiber: OK, resolving this since it's no longer part of this PR.

Simulator/Utils/Global.h
KamathScience: I confirmed that for every opening `#ifdef`, there is a closing `#endif`. For example line 12 - `#ifdef` is indeed closing at line 199.

Also, I am not sure how to go about cleaning this as you correctly pointed we have `#pragma once` and again `#ifdef` directives which are both intended to prevent multiple inclusions of header files. As per C++ standards  `#pragma once` is usually sufficient to prevent multiple inclusions by itself. Maybe I will create a new issue for this ? as there are at least a couple more files that have both `#pragma once` and `#ifdef` directives defined.

stiber: I think it's fine to generate an issue for this. Label it also as "good first issue".

KamathScience: Created a issue #427

stiber (approved):
Looks good! If it's not in the documentation, we might add to our coding style that we use the `= default` notation for constructors and destructors. You can just create an issue for this.

PR #419, opened 2023-03-03
Upgraded GitHub checkout action from v2 to v3

### LINKED ISSUE #417: Upgrade GitHub checkout action to actions/checkout@v2 because Node.js 12 actions are being deprecated
Node.js 12 GitHub actions are being deprecated and should be updated to use Node.js 16. See: [GitHub Actions: All Actions will begin running on Node16 instead of Node12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

In our case, we need to upgrade actions/checkout@v2 to actions/checkout@v3

https://github.com/UWB-Biocomputing/Graphitti/blob/21a4b30fa671b5b5418cb9bafd7cdc0c60e9b87d/.github/workflows/tests.yml#L19

jardiamj:
Closes #417 

Upgrades GitHub checkout action v2 to v3, because actions that use Node.js 12 are being deprecated.


stiber:
Will it be OK to let the eventual merge of the development branch into master fix it there, or is this something that needs to be resolved in master before July?

jardiamj:
> Will it be OK to let the eventual merge of the development branch into master fix it there, or is this something that needs to be resolved in master before July?

it says [Here](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) that "it will come into effect on September 26th" (sic). I assume that by "it" they mean the deprecation of Node 12 actions. So, I believe that we should be OK with letting the fix come into `master` when we merge the development branch.

stiber (approved):

PR #421, opened 2023-04-05
Pointers in HDF5 recorders

xiangli6:
Convert dataset pointers to normal dataset variables in HDF5 recorder classes


stiber (approved):
Looks good!

PR #432, opened 2023-04-10
Create a template for all the factory classes

### LINKED ISSUE #428: Generate factory classes using a template
In the Graphitti codebase, we should consider using a `template` to generate factory classes for `Connections`, `Edges`, `Layout`, `Vertices`, and `Recorders`. As it will avoid duplicating code and ensure consistency in the structure and behavior of our factory classes.

KamathScience:
Fixes #428


Simulator/Utils/Factory.cpp
```diff
@@ -0,0 +1,120 @@
...
+      createFunctionMap["Hdf5GrowthRecorder"] = &Hdf5GrowthRecorder::Create;
+#endif
+   }
+
+   // Register MTRand classes
+   else if constexpr (std::is_same_v<T, MTRand>) {
+      createFunctionMap["MTRand"] = &MTRand::Create;
+      createFunctionMap["Norm"] = &Norm::Create;
+   }
+   return createFunctionMap;
```
stiber, line 103: Is there any way to guard against `T` not matching any of the tests? Or should the empty map be detected by the calling function? By the way, can this whole function be `constexpr`?

Simulator/Utils/Factory.cpp
```diff
@@ -0,0 +1,120 @@
...
+   getCreateFunctionForType<AllVertices>(AllVertices *);
+
+template std::map<std::string, IRecorder *(*)(void)>
+   getCreateFunctionForType<IRecorder>(IRecorder *);
+
+template std::map<std::string, AllEdges *(*)(void)> getCreateFunctionForType<AllEdges>(AllEdges *);
+
+template std::map<std::string, Layout *(*)(void)> getCreateFunctionForType<Layout>(Layout *);
+
+template std::map<std::string, MTRand *(*)(void)> getCreateFunctionForType<MTRand>(MTRand *);
```
stiber, line 120: Why are these needed?

stiber (changes requested):
Looks like this will be a big win...

CMakeLists.txt
```diff
@@ -292,6 +295,7 @@ if(ENABLE_CUDA)
         set_property(TARGET Core PROPERTY CUDA_SEPARABLE_COMPILATION ON)
         set_property(TARGET Connections PROPERTY CUDA_SEPARABLE_COMPILATION ON)
         set_property(TARGET RNG PROPERTY CUDA_SEPARABLE_COMPILATION ON)
+        set_property(TARGET Utils PROPERTY CUDA_SEPARABLE_COMPILATION ON)
```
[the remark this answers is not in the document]
KamathScience: This is redundant and will be deleted. Thanks!

Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include <algorithm>
 
-ConnStatic::ConnStatic()
+ConnStatic::ConnStatic() : Connections()
```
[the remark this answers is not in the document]
KamathScience: Agreed, thought of adding it so it is consistent with the ConnGrowth() constructor.
Shall I remove it from both files?

Simulator/Utils/Factory.cpp
```diff
@@ -0,0 +1,120 @@
...
+#if defined(HDF5)
+   #include "Recorders/Hdf5Recorder.h"
+   #include "Recorders/Neuro/Hdf5GrowthRecorder.h"
+#endif
+
+//MTRand
+#include "RNG/MTRand.h"
+#include "RNG/Norm.h"
+
+template <typename T> std::map<std::string, T *(*)(void)> getCreateFunctionForType(T *placeholder)
```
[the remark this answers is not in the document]
KamathScience: Noted. Will add the documentation

Simulator/Utils/Factory.cpp
KamathScience: By placing the declaration in the header and definition in the cpp file I was guarding instantiating the template against any non-declared type. However, in order to make the code more readable I have removed the explicit instantiation and instead added a static_assert to check for unexpected types.

Simulator/Utils/Factory.cpp
KamathScience: Added a static_assert to check for unexpected types.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+   /// Makes class-to-function map an internal factory member.
+   FunctionMap createFunctions;
+
+   /// Constructor is private to keep a singleton instance of this class.
+   Factory(std::map<std::string, CreateFunction> map)
+   {
+      createFunctions = std::move(map);
+   }
+
+   /// Register class and it's create function to the factory.
```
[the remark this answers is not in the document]
KamathScience: Removed the registerClass method

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+   ///
+   ///  @param  className  class name.
+   ///  @param  function   Pointer to the class creation function.
+   void registerClass(const std::string &className, CreateFunction function)
+   {
+      createFunctions[className] = function;
+   }
```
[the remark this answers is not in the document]
KamathScience: Removed the registerClass method

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
+/**
+ * @file Factory.h
+ * 
+ * @ingroup Simulator/Utils
+ *
+ * @brief A factory template for creating factory class for 
+ *        Connections, Edges, Layout, Verticies, Recorder and MTRand.
```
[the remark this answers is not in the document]
KamathScience: Agreed.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <string>
+
+template <typename T> std::map<std::string, T *(*)(void)> getCreateFunctionForType(T *placeholder);
```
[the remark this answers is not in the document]
KamathScience: Noted, will change the name.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+   {
+      T *placeholder = nullptr;
+      static Factory instance(getCreateFunctionForType<T>(placeholder));
+      return instance;
+   }
+
+   // Invokes constructor for desired concrete class
+   ///
+   /// @param className class name.
+   /// @return Shared pointer to an instance if className is found in
```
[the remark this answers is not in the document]
KamathScience: Will correct this.

Simulator/Utils/Factory.cpp
KamathScience: >  can this whole function be constexpr?

Unfortunately, std::map is not constexpr compliant, however, we can create a template to generate a constexpr map with the help of std::arrays and std::pair, but it might add a lot of complexity to the existing code

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+#include <memory>
+#include <string>
+
+template <typename T> std::map<std::string, T *(*)(void)> getCreateFunctionForType(T *placeholder);
+
+template <typename T> class Factory {
+public:
+   ~Factory() = default;
+
+   static Factory &getInstance()
```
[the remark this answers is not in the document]
KamathScience: Added documentation, please let me know if anything more needs to be added.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,76 @@
...
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <string>
+
+template <typename T> std::map<std::string, T *(*)(void)> getCreateFunctionForType(T *placeholder);
+
+template <typename T> class Factory {
```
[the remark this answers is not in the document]
KamathScience: Added documentation, please let me know if anything more needs to be added.

Simulator/Utils/CreateFunctionMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+// MTRand
+#include "RNG/MTRand.h"
+#include "RNG/Norm.h"
+
+/**
+ *
+ * The templated CreateFunctionMap allows each class to create a concrete
+ * instance of itself at runtime. This approach helps to keep the code organized
+ * and avoids circular dependencies that can make the code less error-prone.
```
stiber, line 56: make the code _more_ error prone?

Simulator/Utils/CreateFunctionMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * How it avoids circular dependency?
+ * If a map was passed from the base class with all the instance of its derived
+ * class, it would create a circular dependency. Meaning, that the derived class
```
stiber, line 60: _classes_

Simulator/Utils/CreateFunctionMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "RNG/MTRand.h"
+#include "RNG/Norm.h"
+
+/**
+ *
+ * The templated CreateFunctionMap allows each class to create a concrete
+ * instance of itself at runtime. This approach helps to keep the code organized
+ * and avoids circular dependencies that can make the code less error-prone.
+ *
+ * How it avoids circular dependency?
```
stiber, line 58: I think, at the start of this (or beforehand), you need to introduce that the factory needs to have a map that includes, for each concrete, derived class its name (string) and the creation function for that particular class. So, then the issue is "in what class do you populate this map, given that it needs to be done before the factory can instantiate any object?" That makes the circular dependency mentioned below clear. The downside (yeh, almost always is one) is that the approach chosen here requires that this factory template needs to include every definition of every class that might be instantiated. There is a plus, which is that all of the map entries are organized in this one file.

Simulator/Utils/CreateFunctionMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+ * undesired type; thereby generating a compile-time error for any request to
+ * instantiate a template for a type that has not been explicitly defined.
+ *
+ * Why the T* argument?
+ * C++ functions resolve polymorphic behavior based on their input arguments,
+ * not their return types. This is because the compiler generates a separate
+ * instance of the function for each function signature. For the templated
+ * CreateFunctionMap, the T* argument is only used to resolve polymorphic
+ * behavior. Also, a pointer is more flexible to pass as a parameter than an
+ * object as the class may or may not have a public constructor and destructor.
```
stiber, line 76: Also, you can create a pointer, `T*`, without having to create an instance of the class `T`. In this case, it is essential, because the function expects a pointer to the (abstract) base class.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,102 @@
...
+ * A static assert is used to check for any undesired type; thereby generating a compile-time error for any request to instantiate 
+ * a template for a type that has not been explicitly defined.
+ * 
+*/
+
+template <typename T> std::map<std::string, T *(*)(void)> CreateFunctionMap(T *placeholder);
+
+
+/**
+ * Factory is one of the important design pattern of Graphitti. It is developed in order to
```
stiber, line 36: patterns

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,102 @@
...
+ * Meaning, that the derived class would need to reference the base class,  
+ * and the base class would also need to reference the derived class for the sake of this map, leading to an infinite loop of references.
+ * 
+ * In addition, the templated CreateFunctionMap safeguards against creating a Factory code of an undesired type.
+ * A static assert is used to check for any undesired type; thereby generating a compile-time error for any request to instantiate 
+ * a template for a type that has not been explicitly defined.
+ * 
+*/
+
+template <typename T> std::map<std::string, T *(*)(void)> CreateFunctionMap(T *placeholder);
```
stiber, line 32: One question: why is this a function and not a static method of the template class?

stiber (changes requested):
There's always a bit more...

Simulator/Utils/Factory.h
KamathScience: Good suggestion, will make this a static method of Factory template.

Simulator/Utils/CreateFunctionMap.cpp
KamathScience: Agreed, and changed accordingly.

Simulator/Utils/Factory.h
stiber: If you do that (and it works), then is it necessary to pass in a parameter at all? The template `T` will already be bound within the method to an actual base class, so I think you can just test it.

Simulator/Utils/Factory.h
```diff
@@ -0,0 +1,225 @@
...
+
+   /// Constructor is private to keep a singleton instance of this class.
+   Factory(std::map<std::string, CreateFunction> map)
+   {
+      createFunctions = std::move(map);
+   }
+
+   static std::map<std::string, T *(*)(void)> CreateFunctionMap(T *placeholder)
+   {
+      std::map<std::string, T *(*)(void)> createFunctionMap;
```
stiber, line 149: `T *(*)(void)` can be replaced with `CreateFunction`. Also, as indicated in a reply to another comment, I think that this method no longer needs a parameter.

BTW, was `Factory.cpp` done away with? I actually think that's OK in this case.

stiber (changes requested):
I think we're close!

Simulator/Utils/Factory.h
KamathScience: It works! I really appreciate you guiding me through this. I learned a lot.

Simulator/Utils/Factory.h
KamathScience: > T *(*)(void) can be replaced with CreateFunction.
 
Agreed.

> was Factory.cpp done away with? 

I was getting linker errors while trying to make the method static, and it would only resolve by adding Factory.cpp in Factory.h file. Creating a circular dependency. So decided to remove it altogether

KamathScience: Any tips on resolving the code style violations? `clang-format -i Simulator/Utils/Factory.h`  does not work but `clang-format -style=llvm -i Simulator/Utils/Factory.h` works, but the changes made are still not accepted by GitHub checks

stiber: Did you try running it manually from the command line? If you do that, you can have it output a modified version of the file that is "fixed". You can then `diff` it with the original to see what it changed (and to check it didn't damage anything, just to be paranoid).

KamathScience: Yeah, I did try it, but unfortunately, it didn't help.

KamathScience: I remember we upgraded the clang-format on our machines to the latest one. But when I check for the version it is still the older version 
```
dkamath@otachi:~/Graphitti$ clang-format -version
clang-format version 7.0.1-8+deb10u2 (tags/RELEASE_701/final)

```

stiber: The latest version is available on `csslab13.uwb.edu`.

KamathScience: Thank you this suggestion helped!

stiber (approved):
Great job!

PR #463, opened 2023-06-19
[issue #406] Removed the unnecessary Connections parameter level from XML config file

### LINKED ISSUE #406: Connections parameters for ESCS simulations shouldn't have a extra level in the XML config file
https://github.com/UWB-Biocomputing/Graphitti/blob/b5bd8d3298e67b620f96032cfb5c7f484345ec7e/configfiles/test-king-county-911.xml#L34-L40

jardiamj:
Closes #406 

Removes the extra level in the Connections parameter node from the XML config file because it is not needed.

### DIFF
--- a/configfiles/test-king-county-911.xml
+++ b/configfiles/test-king-county-911.xml
@@ -31,13 +31,11 @@
     </EdgesParams>
 
     <ConnectionsParams class="Connections911" name="ConnectionsParams">
-      <Connections911Params name="Connections911Params">
-        <graphmlFile name="graphmlFile">../configfiles/graphs/King_county_NG911.graphml</graphmlFile>
-            <!-- Max ConnsPerVertex = 87 RC + 1 PR -->
-        <connsPerVertex name="connsPerVertex">100</connsPerVertex>
-        <psapsToErase name="psapsToErase">1</psapsToErase>
-        <respsToErase name="respsToErase">2</respsToErase>
-      </Connections911Params>
+      <graphmlFile name="graphmlFile">../configfiles/graphs/King_county_NG911.graphml</graphmlFile>
+          <!-- Max ConnsPerVertex = 87 RC + 1 PR -->
+      <connsPerVertex name="connsPerVertex">100</connsPerVertex>
+      <psapsToErase name="psapsToErase">1</psapsToErase>
+      <respsToErase name="respsToErase">2</respsToErase>
     </ConnectionsParams>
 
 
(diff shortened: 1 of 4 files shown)


configfiles/test-king-county-911.xml
```diff
@@ -31,13 +31,11 @@
...
-      <Connections911Params name="Connections911Params">
-        <graphmlFile name="graphmlFile">../configfiles/graphs/King_county_NG911.graphml</graphmlFile>
-            <!-- Max ConnsPerVertex = 87 RC + 1 PR -->
-        <connsPerVertex name="connsPerVertex">100</connsPerVertex>
-        <psapsToErase name="psapsToErase">1</psapsToErase>
-        <respsToErase name="respsToErase">2</respsToErase>
-      </Connections911Params>
+      <graphmlFile name="graphmlFile">../configfiles/graphs/King_county_NG911.graphml</graphmlFile>
+          <!-- Max ConnsPerVertex = 87 RC + 1 PR -->
+      <connsPerVertex name="connsPerVertex">100</connsPerVertex>
```
stiber, line 36: Is this a max or an exact number? If the former, name it `maxConnsPerVertex`.

stiber (changes requested):
Does this need a code change? One other question.

configfiles/test-small-911.xml
```diff
@@ -37,7 +37,7 @@
       <Connections911Params name="Connections911Params">
         <graphmlFile name="graphmlFile">../configfiles/graphs/small_911.graphml</graphmlFile>
             <!-- Max ConnsPerVertex = 87 RC + 1 PR -->
-        <connsPerVertex name="connsPerVertex">88</connsPerVertex>
+        <connsPerVertex name="maxConnsPerVertex">88</connsPerVertex>
```
stiber, line 40: Change needs to be made to both the tag name and the `name` attribute. Actually, the latter is intended to be used by Workbench in its GUI, so it should really be more user friendly, such as `name="Maximum Connections per Vertex"`

Also, the same change needs to be made to the other config file.

stiber (changes requested):

configfiles/test-small-911.xml
jardiamj: I assume that we have to make all those `name` attributes in the config files more user-friendly. I also foresee that these config files will change as I complete the ESCS model implementation.

stiber: I agree.

stiber (approved):

PR #486, opened 2023-07-27
[ISSUE-410] Check switches on edgeType and make sure they have default cases

### LINKED ISSUE #410: Check switches on edgeType and make sure they have default cases
This adds no overhead to runtime but would catch when something didn't update from its initialization state.
See also issue #409 and #404

Andrzej-Dawiec:
Closes #410 
This PR checks to make sure switches on edgeType and makes sure they have default cases, with the ability to expand with throwing exceptions in the future if the switches return 0. In principle, return 0 in switch case if error.


Andrzej-Dawiec:
Had trouble with comment indentation but it should be good now.

stiber (approved):

PR #498, opened 2023-08-08
[ISSUE 464] Add cluster point process call generation script

### LINKED ISSUE #464: Add cluster point process call generation script
The call arrival process is being decoupled from the Simulator through the use of an external input file, that provides the list of calls to be processed during the simulation. This will allow us to experiment with different call arrival models.

A spatiotemporal cluster point process provides a reasonable framework for modeling call arrivals. Therefore, we will script the generation of calls into a .xml file using this framework.

jardiamj:
Closes #464

This PR adds a python script under `Tools` for modeling the call arrival process as a cluster point process.


jardiamj:
Closing PR because I accidentally opened it against the `master` branch instead of `development`.

PR #524, opened 2023-09-05
[ISSUE 499] review spd graph

### LINKED ISSUE #499: Review Seattle PD PSAP graphml file for correctness
The graphml containing the Seattle PD PSAP graph was extracted from the King county 911 network graph. The King county graph was in turn generated using the Geographic Information System (GIS) files provided by our contacts at Washington State 911 Coordination Office.

The GIS files used for generating the graph are based on service/jurisdictional boundaries for the PSAP and Responder entities. Due to many service boundaries not being contiguous, the generated graph appears to have more responder vertices than there are in reality. For instance, the Seattle Fire Department has 41 locations some of which provide EMS only service, Fire only services, or both; but when examining the generated graph, it contains 55 Seattle FD EMS nodes and 55 Seattle FD Fire nodes. This is consistent with the Fire and EMS layers provided in the GIS files but does not correspond to the location of the Fire Stations nor with the services provided (EMS or Fire).

A more accurate approach would be to use the locations and service type provided by the [Seattle Fire Department website](https://www.seattle.gov/fire/about-us/fire-stations). The [wikipedia page for Seattle FD](https://en.wikipedia.org/wiki/Seattle_Fire_Department) also contains information about each of the stations' location and their units.

Similarly, the location of the Seattle PD stations can be obtained from [their website](https://www.seattle.gov/police/about-us/police-locations).

This can be a tedious work if we have a large network. It would be great if we can find a central source for this data.

jardiamj:
closes #499 

This PR updates the graphml file for the Seattle PD simulation with the actual coordinates of the [33 Fire Stations](https://www.seattle.gov/fire/about-us/fire-stations) and [5 Police Precincts](https://www.seattle.gov/police/about-us/police-locations) that serve the Seattle PD PSAP boundaries. It also modified the scripts that generate the input files to match the caller region's vertex_id.

The file with the manual modifications is `Tools\gis2graph\graph_files\spd.graphml`; the same file is copied under `configfiles\graphs`.


Tools/InputGeneration/ClusterPointProcess/cluster_point_process.py
```diff
@@ -237,7 +237,7 @@ def add_vertex_events(node, vertex_id, vertex_name, data):
 
     # Get the grid for the Seattle PD Caller Region from the graphml file
     graph_file = '../../gis2graph/graph_files/spd.graphml'
-    spd_cr_id = '194'
+    spd_cr_id = '74'
     graph = nx.read_graphml(graph_file)
     spd_grid = np.array(eval(graph.nodes[spd_cr_id]['segments']))
```
stiber, line 242: This suggests there are a number of magic numbers and hard-coding of things in this code. Do you want to add on as a subsequent issue to make this code more general? That could be a good mini-project for an undergrad.

stiber (approved):
Looks fine, as far as I can tell. I think there is a follow-on issue you can create as a "good first issue".

Tools/InputGeneration/ClusterPointProcess/cluster_point_process.py
jardiamj: Yes, it has magic numbers, and I will create an issue.

PR #525, opened 2023-09-08
[Issue 489] Implement call abandonment modeled with patience time

### LINKED ISSUE #489: Implement call abandonment in NG911 model
We will model call abandonment using a patience parameter as used in the Erlang A queuing model. The following is a general description of the process as presented in the literature.

Gans et al. (2003) used the Erlang A queuing model to demonstrate that call abandonment in heavily loaded call centers significantly affects system performance. Abandonment is modeled through a `patience` parameter which is an exponentially distributed random variable with an abandonment rate θ (average patience of 1/θ).

Mandelbaum and Zeltyn (2005) suggest two methods for estimating the average patience (1/θ). The practical approach for the 911 model is to calculate the abandonment rate with the following formula:
`θ = %Abandonment / Average Wait`

Each call gets an exponentially distributed patience time. If the call has not been picked up once its patience runs out, the customer abandons the waiting queue.

References:

Gans, N., Koole, G., & Mandelbaum, A. (2003). Telephone Call Centers: Tutorial, Review, and Research Prospects. Manufacturing & Service Operations Management, 5(2), 79–141. https://doi.org/10.1287/msom.5.2.79.16071

Mandelbaum, A., & Zeltyn, S. (2007). Service Engineering in Action: The Palm/Erlang-A Queue, with Applications to Call Centers. In D. Spath & K.-P. Fähnrich (Eds.), Advances in Services Innovations (pp. 17–45). Springer. https://doi.org/10.1007/978-3-540-29860-1_2

jardiamj:
Closes #489

Implementation of call abandonment using a patience time parameter as in the Erlang A Queueing model.

Call abandonment and average patience estimations are described in the following papers:

Gans, N., Koole, G., & Mandelbaum, A. (2003). Telephone Call Centers: Tutorial, Review, and Research Prospects. Manufacturing & Service Operations Management, 5(2), 79–141. https://doi.org/10.1287/msom.5.2.79.16071

Mandelbaum, A., & Zeltyn, S. (2007). Service Engineering in Action: The Palm/Erlang-A Queue, with Applications to Call Centers. In D. Spath & K.-P. Fähnrich (Eds.), Advances in Services Innovations (pp. 17–45). Springer. https://doi.org/10.1007/978-3-540-29860-1_2


Simulator/Vertices/NG911/All911Vertices.h
```diff
@@ -81,6 +81,8 @@ class All911Vertices : public AllVertices {
    vector<vector<uint64_t>> logAnswerTime_;
    /// The end time for every call in the log
    vector<vector<uint64_t>> logEndTime_;
+   /// Logs if the call was abandoned
+   vector<vector<bool>> logWasAbandoned_;
```
stiber, line 85: Are these data members meant to hold a log of something? Right now, starting their name with `log`, they sound like verbs -- like they produce a log. If they're meant to be a log, then it might be better to end with `log`, like `answerTimeLog_`. OTOH, `log` overlaps with our logging framework. Are they meant to store a history? Then it would be consistent with the nomenclature in our recorders to name them like `answerTimeHistory_`.

stiber (changes requested):

Simulator/Vertices/NG911/All911Vertices.h
jardiamj: My thought was that these represent the Call Log, but now that you point it out they sound like Yoda talk. And I agree that `History` is a better nomenclature so they don't get confused with our logging framework.

stiber (approved):

PR #534, opened 2023-09-20
[ISSUE 490] responder dispatch

### LINKED ISSUE #490: Implement NG911 responder dispatch
We need to implement the responder dispatch logic. The responder that is available, appropriate for, and closest to the emergency event is the one that is dispatched.

In order to implement this we need to model the emergency call types, in addition to time and location. Obtaining 911 data to link the calls to the emergency event types is proving more difficult than originally thought.

### LINKED ISSUE #533: Waiting queue capacity in 911 vertices causes abnormal number of dropped calls for some vertices
The queue capacity in event queueing models is calculated as `trunks - agents`, but this approach causes an issue in our implementation, where some vertices could have an abnormal number of dropped calls because all calls are being queued up before assigning them to an available agent. The solution is to size the capacity of the queue to the number of trunks and keep track of the number of busy agents at every time step.

jardiamj:
Closes #490 
Fixes #533

Implements the basic logic for finding the responder closest to the emergency event and assigns the call to it. It also fixes a bug where the waiting queue was not accounting for the fact that we queued up every call, therefore we were getting more dropped calls than we should.

At the moment, the responder only logs that it received a call. The logic for the responder queue and response time will be implemented in a separate issue.

I also refactored the code in `advanceVertices` so that the logic is divided into `advancePSAP`, `advanceCALR`, and `advanceRESP`. Lastly, I renamed `agents` to `servers` to avoid confusion with agent-based parallel computing.


Simulator/Connections/NG911/Connections911.cpp
stiber, line 1: At this point, we need to add comprehensive class documentation to each of the classes here. They should be formatted to work with Doxygen; I can't recall offhand whether that means they should only be in the `.h`.

Simulator/Connections/NG911/Connections911.h
```diff
@@ -73,6 +72,14 @@ class Connections911 : public Connections {
    ///  @return true if successful, false otherwise.
    virtual bool updateConnections(AllVertices &vertices) override;
 
+   /// Finds the outgoing edge from the given vertex to the Responder closest to
+   /// the emergency call location
+   ///
+   /// @param call         The call that needs a Responder
+   /// @param vertexIdx    The index of the vertex serving the call (A PSAP)
+   /// @return    The index of the outgoing edge to the closest Responder
+   BGSIZE getEdgeToClosestResponder(const Call &call, const BGSIZE vertexIdx);
```
stiber, line 81: Should this be `private`?

Simulator/Edges/NG911/All911Edges.cpp
stiber, line 1: Ditto on the comprehensive class documentation.

Simulator/Layouts/NG911/Layout911.cpp
stiber, line 1: Ditto on the comprehensive class documentation.

Simulator/Utils/Global.h
stiber, line 1: Check to see if we have an issue to clean up this file, deciding if items in it should be moved elsewhere and generally modernizing code. Good for an undergrad.

Simulator/Vertices/NG911/All911Vertices.cpp
stiber, line 1: Ditto on the comprehensive class documentation.

stiber (changes requested):

Simulator/Connections/NG911/Connections911.h
jardiamj: This method is used by `All911Vertices::advanceVertices` to dispatch the appropriate responder. I decided to put it in `Connections911` because that is the class that owns the edges and edgeIndexMap. The other option was to place it in `All911Vertices` passing references to the edges and edgeIndexMap along with the call and vertex index.

jardiamj: I gave this some thought and this is still what makes the most sense to me. And in this case, it needs to be a public method.

stiber (approved):

Simulator/Connections/NG911/Connections911.h
stiber: OK, that's a good reason.

PR #537, opened 2023-09-25
[ISSUE 532] Clean ununsed variables, methods, and code blocks in NG911 model

### LINKED ISSUE #532: Clean up old code from NG911 models
In the previous version of the NG911 models, parameters such as the number of callers, dispatchers, and responders were coded into the simulation. These parameters should now be part of the Graphml configuration file, and therefore this code is no longer needed.

jardiamj:
Closes #532 

Cleans up old 911 code no longer being used, either because the parameters are now part of the Graphml input file or because the model has changed completely.


stiber (dismissed):
Just to double-check, these were code snippets in classes that are still being retained, right?

jardiamj:
> Just to double-check, these were code snippets in classes that are still being retained, right?

That's right. This was commented-out code because it was copied from the Neuro classes, and code that used a vertex grid with quadrants to crate the 911 nodes and edges.

jardiamj:
I had to resolve a merging conflict, which forced me to request another review.

stiber (approved):

PR #539, opened 2023-09-26
[ISSUE 180] Moved width_ and height_ parameter from Simulator to FixedLayout

### LINKED ISSUE #180: move width_ and height_ to layout from simulator
height_ and width_ and their accessors are currently in simulator and should be in layout. this requires a fair amount of refactoring in many files so buckle up! 

and changing all configfiles.

jardiamj:
Closes #180 

Moved the `width_` and `height_` parameters from `Simulator` to `FixedLayout`. The method in `Model` that printed debugging layout output was also moved to `FixedLayout` to conserve it, although it is currently not being used.


Simulator/Layouts/Neuro/FixedLayout.cpp
```diff
@@ -129,11 +130,10 @@ void FixedLayout::loadParameters()
...
    }
 
-   // Get the total number of vertices
-   int width, height;
-   ParameterManager::getInstance().getIntByXpath("//PoolSize/x/text()", width);
-   ParameterManager::getInstance().getIntByXpath("//PoolSize/y/text()", height);
-   numVertices_ = width * height;
+   // Get width, height and total number of vertices
+   ParameterManager::getInstance().getIntByXpath("//PoolSize/x/text()", width_);
+   ParameterManager::getInstance().getIntByXpath("//PoolSize/y/text()", height_);
```
stiber, line 135: I think that we should move these parameters to the `<LayoutParams class="FixedLayout" name="LayoutParams">` section of the config files. If you like, create a new issue to do this, with a note of where in this file the change is needed (and, of course, that all the config files need changing).

Simulator/Layouts/Layout.cpp
```diff
@@ -58,6 +58,11 @@ void Layout::registerGraphProperties()
    // initial graph
 }
 
+void Layout::printLayout()
+{
+   // Empty implementation because not all Layouts implement it but it is used by the Model
+}
+
```
stiber, line 65: Can't this be pure virtual in this class?

stiber (changes requested):

Simulator/Layouts/Layout.cpp
jardiamj: > Can't this be pure virtual in this class?

Currently, the code in this method is never executed because the call in `Simulator::simulate()`, where it was being used for debugging, is commented out:

```
   // DEBUG_MID(model->logSimStep();) // Generic model debug call
```

The code seems to print a formatted grid of the neurons' layout to standard output, so I moved it to `FixedLayout` but kept it as a call to `Layout` in `Model` without down-casting. And, since no other Layout implements this I had to add an empty implementation in the base class.

This is a hack to keep a block of code that is not even being used, so I don't know if it is worth it. I believe this was used at some point for debugging the `Connection Growth` simulation model but is now dead code left behind. We could remove the `Model::logSimStep()` method and leave the piece code snippet in `FixedLayout` only, in case anyone ever needs to use it.

stiber: OK, I say comment out all of this dead code, with a brief comment.

jardiamj: I commented out the dead code in the Model class with a comment, removed the method from the Layout base class, and put a comment in the FixedLayout class where I moved the original code snippet, along with the `width_` and `height_` variables.

Simulator/Layouts/Neuro/FixedLayout.cpp
jardiamj: Created a new issue for this.

stiber (approved):

PR #551, opened 2023-10-27
Issue 548 visualizer file loader

### LINKED ISSUE #548: Add file loader dialog box to Visualizer

whiteja2:
Closes #548

### DIFF
--- a/Tools/graphMLVis/visualize_graphML.py
+++ b/Tools/graphMLVis/visualize_graphML.py
@@ -6,9 +6,73 @@ import numpy as np
 from shapely.ops import unary_union
 from shapely.geometry import Polygon
 import math
+import os
+from tkinter import *
+from tkinter import filedialog
+
+# start of dialog box
+window = Tk()
+
+# variables acquired through dialog box
+class Paths:
+    def __init__(self, graphml_path, node_type):
+        self.graphml_path, self.node_type = graphml_path, node_type
+
+    def changegraphml(self, path):
+        self.graphml_path = os.path.normpath(path)
+
+    def changenodetype(self):
+        self.node_type = radio_var.get()
+
+dVars = Paths(os.path.normpath("Tools\gis2graph\graph_files\King_county_NG911.graphml"), "EMS")
+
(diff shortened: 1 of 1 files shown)


stiber:
I note that there are file paths written "Windows style", using back slashes as separators. These are wrapped in calls to `os.path.normpath()`. Is that common practice in Python? In most languages, the backslash in strings is used as an escape character. Just checking before I approve.

jardiamj:
In Linux, the backslash will be identified as an escape character therefore path strings with backslashes only work on Windows. I suggested to him, a few days ago, the use of `os.path` to make the script OS-agnostic but what I had in mind was to assemble the path like this:

```python
os.path.join('Tools','gis2graph','graph_files','King_county_NG911.graphml')
```

whiteja2:
Based on what I found online, os.path.normpath() converts the slashes to
the appropriate form based on which operating system is being used. Through
testing, I know for certain that it can convert forward slashes to
backslashes for my Windows machine and assumed the inverse would be true. I
can go about converting all of the hardcoded paths to use forward slashes
for consistency in the code!

For the dialog box, I found os.path.normpath() to be a simpler solution as
the user passes a file path and it's then converted to the proper format
without the need for parsing the string.

stiber (approved):
Looks good! I think forward slashes are more likely to not cause problems or make the reader wonder.

jardiamj:
My bad, I had only tested the previous version. @whiteja2 is correct, `os.path.normpath` does convert the path to forward slashes in Linux. After trying this new version, it does work as intended.

stiber:
> My bad, I had only tested the previous version. @whiteja2 is correct, `os.path.normpath` does convert the path to forward slashes in Linux. After trying this new version, it does work as intended.

I was mostly concerned with the string itself. In other languages, the backslash has a special meaning when used in a string and you need `"\\"` if you want a literal backslash. So, if a forward slash will work here, then using them is preferable, as it helps code readability (by avoiding requiring the reader to either know that Python doesn't care when used in strings or requiring them to look it up).

PR #556, opened 2023-11-05
[ISSUE 538] Add recording of queueLength and utilization histories for NG911

### LINKED ISSUE #538: Record NG911 system performance history
We need to record system performance metrics for the 911 queueing model. There are two measurements that allow us to weigh the load of the system:

- **System utilization:** The proportion of the time during the simulation that the servers are busy.
- **Time Average Queue Length:** The time-weighted average number of customers in the queue during the simulation.

These measurements can be estimated from the history of the number of busy servers at each time step, and the history of the queue length at each time step; respectively.

jardiamj:
Closes #538 

This PR adds the recording of queue length and utilization histories for the NG911 PSAP and Responder nodes.
These histories are recorded on a per-time-step basis.

In addition, I changed the average on-scene time used in the cluster point process script and the Seattle PD real data, to 20 minutes. This change was based on some research papers, which are described as comments in the scripts.


Simulator/Vertices/NG911/All911Vertices.cpp
```diff
@@ -65,7 +70,7 @@ void All911Vertices::createAllVertices(Layout &layout)
       assert(*vi < size_);
```
stiber, line 70: Is this code CPU-only (for example, used only for loading graphs during initialization) or is it something that would also be implemented in the future on the GPU? If the former case, then this is fine; no change needed. If the latter case, then we want to avoid libraries not available on the GPU and we want to use indices into vectors, rather than vertex pointers. I suspect the former, but I want to double-check.

Simulator/Vertices/NG911/All911Vertices.h
```diff
@@ -138,14 +138,19 @@ class All911Vertices : public AllVertices {
    /// True if the call was abandoned
    vector<vector<bool>> wasAbandonedHistory_;
```
stiber, line 139: Is this a variable that will be needed on the GPU? If so, make it `vector<vector<unsigned char>>` and treat its elements as having `bool` values (i.e., using `true` and `false`).

stiber (changes requested):
Might not be changes...

Simulator/Vertices/NG911/All911Vertices.cpp
jardiamj: Yes, this is CPU-only code executed during initialization for loading the graph into Graphitti's data structures.

stiber (approved):

Issue #459, opened 2023-06-12
Serialization and GPU Compatibility of Boolean Values in Graphitti


KamathScience:
#### Description
Currently, Graphitti utilizes a dynamically allocated bool array `inUse_` in AllEdges to store boolean values. 
` unique_ptr<bool[]> inUse_;`

However, Cereal Library does not serialize or deserialize dynamically allocated arrays. So, an alternative option considered was using a vector of bool, which provides a compact representation using a single bit per element to conserve memory. However, this representation is not suitable when transferring the data to the GPU, as the GPU expects boolean values in the form of true or false rather than compact 0s and 1s.

To address this issue, a potential solution is to convert the `inUse_` to vectors of int or unsigned char. This conversion would allow for a compatible data representation that can be serialized using Cereal Libray and data can be transferred to the GPU.

PR #564, opened 2023-12-01
[ISSUE-459] Change vector of bool to vector of unsigned char

KamathScience:
Closes issue #459

This pull request updates the `inUse` member variable within the `AllEdges` class, shifting from `std::vector<std::unique_ptr<bool>>` to `std::vector<unsigned char>`. The decision to utilize `unsigned char` for storing Boolean values is driven by a focus on significantly enhancing serialization. This enables consistent use of the `serialize` method, thereby streamlining the codebase and eliminating the need for separate `load` and `save` methods in the cereal library.

Opting for `unsigned char` brings about a more memory-efficient solution compared to `int`. Additionally, it distinguishes itself from other member variables that store char.

 - [x] I have tested this change for GPU

### DIFF
--- a/Simulator/Connections/Neuro/ConnGrowth.cpp
+++ b/Simulator/Connections/Neuro/ConnGrowth.cpp
@@ -254,7 +254,7 @@ void ConnGrowth::updateSynapsesWeights()
          BGSIZE synapse_adjusted = 0;
          BGSIZE iEdg = Simulator::getInstance().getMaxEdgesPerVertex() * destVertex;
          for (BGSIZE synapseIndex = 0; synapse_adjusted < synapseCounts; synapseIndex++, iEdg++) {
-            if (synapses.inUse_[iEdg] == 1) {
+            if (synapses.inUse_[iEdg]) {
                // if there is a synapse between a and b
                if (synapses.sourceVertexIndex_[iEdg] == srcVertex) {
                   connected = true;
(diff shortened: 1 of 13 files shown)


Simulator/Connections/Neuro/ConnGrowth.cpp
```diff
@@ -254,7 +254,7 @@ void ConnGrowth::updateSynapsesWeights()
          BGSIZE synapse_adjusted = 0;
          BGSIZE iEdg = Simulator::getInstance().getMaxEdgesPerVertex() * destVertex;
          for (BGSIZE synapseIndex = 0; synapse_adjusted < synapseCounts; synapseIndex++, iEdg++) {
-            if (synapses.inUse_[iEdg] == true) {
+            if (synapses.inUse_[iEdg] == 1) {
```
stiber, line 257: I hate to be "that guy", but I can't help it. Is there a standard for conversion between `unsigned char` and the `const bool` values `true` and `false`? (Well, clearly `false` converts back and forth to `0`; it has to.) Would it be better to use `true` and `false`, or is it better to use `1` and `0`, so there's no confusion about the base type? On the other hand, using the number could lead to misinterpretation when reading a line like this that `inUse` is a count, rather than a flag.

As they say, the lower the stakes the greater the controversy.

stiber (changes requested):
Might not be a change.

Simulator/Connections/Neuro/ConnGrowth.cpp
KamathScience: Agreed, using 1 and 0 might be confusing. Switching to 'true' and 'false' improves readability, so I'll stick with that for clarity.

stiber (approved):

PR #573, opened 2023-12-10
[ISSUE-502]  add an abstract base class ,RecordableBase defining the interface for recordable variables

### LINKED ISSUE #502: Create a recordable template for EventBuffer class to enable XmlRecorder works with recordable
Create a recordable template for EventBuffer class and modify the code in the XmlRecorder

Currently, the EventBuffer only stores uint64_t data, but there is a need for the simulator to store other types of data such as double,  To solve this problem, we need to create a recordable template for the EventBuffer class and modify the code in the XmlRecorder so that it can work with Recordable.

xiangli6:
Closes #502 

This pull request introduces a Recordable template for the EventBuffer class, incorporates a RecordableBase interface for the Recordable template, and adjusts the XmlRecorder code accordingly. The C++17 std::variant feature is utilized in the Recorder module to manage a list of different data types for recorded variables. Additionally, the unit tests for XmlRecorder have been revised to accommodate these changes.

The RecordableBase class serves as the base class for all variables that need to be recorded during simulation. It defines the interface for accessing information about recorded variables. 

The Recordable class template extends the `RecordableBase` interface and provides a concrete implementation for recording variables during simulation. It is templated to accommodate different data types for recording.


Simulator/Recorders/Hdf5Recorder.cpp
```diff
@@ -487,11 +487,11 @@ void Hdf5Recorder::printParameters()
                                    << "\tResult file path: " << resultFileName_ << endl);
 }
 
-void Hdf5Recorder::registerVariable(std::string varName, EventBuffer &recordVar)
+void Hdf5Recorder::registerVariable(std::string varName, RecordableBase *recordVar)
 {
 }
 
-void Hdf5Recorder::registerVariable(string varName, vector<EventBuffer> &recordVar)
+void Hdf5Recorder::registerVariable(string varName, vector<RecordableBase *> recordVars)
```
stiber, line 494: Are we using raw pointers? [Note, I'll only comment here, but of course this applies in multiple places.] This could be a special case, since these are literally pointers to member variables of other classes/objects and dynamic memory allocation, etc. is irrelevant (and I think we talked about this already, but after a long day of grinding emails, I can't be sure). Hey @KamathScience , do we have anything written up about how we use smart pointers and where raw pointers are permissible? Thoughts?

Simulator/Recorders/Hdf5Recorder.h
```diff
@@ -70,9 +70,9 @@ class Hdf5Recorder : public Recorder {
    /// Registered to OperationManager as Operation::printParameters
    virtual void printParameters() override;
 
-   virtual void registerVariable(std::string varName, EventBuffer &recordVar) override;
+   virtual void registerVariable(std::string varName, RecordableBase *recordVar) override;
 
-   virtual void registerVariable(std::string varName, vector<EventBuffer> &recordVar) override;
+   virtual void registerVariable(std::string varName, vector<RecordableBase *> recordVars) override;
```
stiber, line 75: Methods need Doxygen comments. If you want to defer this, make an issue and respond/resolve this.

Simulator/Recorders/Recorder.h
```diff
@@ -21,19 +23,22 @@ class Recorder {
...
+   // ToDo : remove it ?
    // Get the current radii and rates values
    virtual void getValues() = 0;
 
    // Terminate process
    virtual void term() = 0;
 
-   // Compile history information in every epoch
+   // Compile/capture history information in every epoch
    // @param[in] neurons   The entire list of neurons.
```
stiber, line 42: Comment needs modification, to make it simulation domain agnostic.

docs/Developer/ClassDiagrams/recorder.puml
stiber, line 1: Should this file name be capitalized?

docs/Developer/ClassDiagrams/vertices.puml
stiber, line 1: Should this file name be capitalized?

stiber (changes requested):
Mostly nitpicks and commenting requests. Two questions, with potential for discussion with others.

Simulator/Recorders/Hdf5Recorder.cpp
xiangli6: we previously discussed that while raw pointers cannot be used as data members in Graphiit, they can be utilized as parameters in a method. 

To address this, my current approach involves passing a reference of the recorded variable through a pointer in the registerVariable method. Within this method, the pointer is dereferenced, and the reference is then passed to a shared pointer in the variable table. This approach aims to avoid mixing raw pointers and shared pointers, as combining ordinary pointers with smart pointers can lead to double free problems. 

If the use of raw pointers as parameters is not permitted, I propose using a reference instead of a raw pointer or smart pointer in the registerVariable method, and removing the overloaded function registerVariable(string varName, vector<RecordableBase *> recordVars). The use of a reference prevents the addition of the overloaded function registerVariable(string varName, vector<RecordableBase> &recordVars) due to the conversion issue from vector<EventBuffer> to vector<RecordableBase. This suggestion is based on the need for the recorder to register the address for recorded variables that are already dynamically allocated, such as vertexEvent_, a data member of the Vertice object allocated during the simulation run, and should not be owned by the recorder. Since the recorder should not assume ownership of these variables, and the objects will not be deleted when the recorder finishes recording values, a shared pointer is not suitable in this case, especially if we use the make_shared function. The use of a constructor (such as code in XmlRecorder.h variableLocation_ = shared_ptr<RecordableBase>(&location, [](RecordableBase *) {
            });) instead of make_shared when registering a variable is one potential solution for utilizing shared_ptr.

KamathScience: Hey @stiber @xiangli6, we do have a guideline in our C++ coding standards that recommends passing parameters as references wherever possible. If a parameter is not modified, pass it as a `const` reference, otherwise just pass it as a reference.

So the function could be modified as follows:

`void Hdf5Recorder::registerVariable(const std::string& varName, RecordableBase& recordVar)`


`void Hdf5Recorder::registerVariable(const std::string& varName, std::vector<RecordableBase*>& recordVars)`


Here, both the `varName` and the `recordVars` parameters are passed as  references to avoid unnecessary copying. Additionally, the RecordableBase objects are passed by reference, and the vector is a vector of pointers to RecordableBase (storing pointers does not hinder the readability, but let's consider storing references).

Few nitpicks:
- Always use `const` for variables that are not modified in the function.
- If using, let's consistently use `std::` prefix for standard library types else none at all.

Simulator/Recorders/Recordable.h
```diff
@@ -0,0 +1,61 @@
...
+ * about the recorded variable.
+ */
+
+#pragma once
+#include "RecordableBase.h"
+#include <vector>
+
+template <typename T> class Recordable : public RecordableBase {
+protected:
+   /// Holds the event time steps
```
KamathScience, line 18: Let us move the protected section after the public section

Simulator/Recorders/Recordable.h
```diff
@@ -0,0 +1,61 @@
...
+   // @brief Start a new epoch for the recordable variable.
+   virtual void startNewEpoch()
+   {
+      return eventTimeSteps_.clear();
+   }
+
+   // @brief Get the value of the recordable variable at the specified index.
+   // return A variant representing the recorded value (uint64_t, double, or string)
+   virtual std::variant<uint64_t, double, string> getElement(int index) const
+   {
```
KamathScience, line 51: I understand this function is a implementation of the virtual method. However, using a variant might be hindering readability in this specific context. The vector stores type `T`, and it seems the function essentially returns a value from the vector. Simplifying the return type to `T` could enhance code clarity.

Simulator/Recorders/Recordable.h
```diff
@@ -0,0 +1,61 @@
...
+      return eventTimeSteps_[index];
+   }
+
+   // @brief Get A string representing the data type of the recordable variable
+   virtual string getDataType() const
+   {
+      return basicDataType_;
+      ;
```
KamathScience, line 59: nitpick: extra `;`

Simulator/Recorders/Recordable.h
```diff
@@ -0,0 +1,61 @@
...
+   }
+};
```
KamathScience, line 61: Please add `override` keyword where necessary

Simulator/Recorders/XmlRecorder.h
```diff
@@ -69,60 +84,91 @@ class XmlRecorder : public Recorder {
       resultFileName_ = fileName;
    }
 
-   // Accessor method for neuronName_ (only included during unit tests)
+   // Accessor method for variable name (only included during unit tests)
    // @param numIndex   The index number in the variable list.
    std::string getVariableName(int numIndex) const
    {
       return variableTable_[numIndex].variableName_;
    }
```
KamathScience, line 92: If possible pass string and other object as a reference (or const reference) from getter methods.

Simulator/Recorders/Hdf5Recorder.cpp
stiber: Ah, is that in https://github.com/UWB-Biocomputing/Graphitti/blob/development/docs/Developer/cppStyleGuide.md? Maybe we need to do a "mini release" where we cherry pick documentation changes into the master branch.

stiber: So, the issue becomes that the Recorders need to assemble what is (simplified version) a vector of references to member variables in other objects. This isn't directly possible, but there is a [`std::reference_wrapper`](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) that enables such a thing.

So, question: is this desirable or is it unnecessarily complex? I think we have a basic principle that we can use raw pointers when they are ephemeral, for example, local variables that go away when a method returns. In this case, these are pointers that are retained in the member variables of a class. However, I believe (?) there is no intent for these member variables to be serialized, since serialization happens at the end of a simulation (and all output files will also be finalized at that point) and deserialization happens at the beginning of a simulation (and all newly created simulation objects will register whatever variables they like to the Recorder). Thoughts?

KamathScience: +1.  `std::reference_wrapper` can indeed get complex, so yeah raw pointer is a better choice here (its simple and it does not affect serialization).

Also, we have a section in the cppStyleGuide:

> Inputs and Outputs
> - Parameters are either inputs to the function, outputs from the function, or both. Non-optional input parameters should usually be values or const references, while non-optional output and input/output parameters should usually be references (which cannot be null).

Maybe I need to add more clarity in our guide?

stiber: Probably good to first make it clear that this is not about `cout`/`cerr`, but rather function parameters/return values. The content is hard to nuance. We want to avoid passing raw pointers. Smart pointers could make sense, but references are simpler. So, maybe it needs to be:

1. Use values/references/const references wherever it is straightforward.
2. If using the above is problematic, prefer smart pointers.
3. If the above is complex, we can use raw pointers. Document the use carefully in comments, to explain the decision. Also, before making this choice, check that the pointers are either ephemeral (will not be stored beyond the execution lifespan of the function) or at least will never need to be serialized. Indicate this in the comment.

Simulator/Recorders/XmlRecorder.h
```diff
@@ -23,11 +23,9 @@
 #include "Global.h"
 #include "Model.h"
 #include "Recorder.h"
-#include <fstream>
-#include <vector>
 
-// a list of basic data types in different recorded variables
-typedef std::variant<uint64_t, double, string> multipleTypes;
+/// a list of basic data types in different recorded variables
+typedef variant<uint64_t, double, string> multipleTypes;
```
stiber, line 28: Better to use `using`.

stiber (changes requested):
Just one more thing (yes, I am annoying).

KamathScience (commented):
Good work, thank you Xiang

stiber (approved):

PR #583, opened 2024-01-03
[ISSUE-510] Remove redundant file

### LINKED ISSUE #510: Remove redundant StudentSetup.md file
Graphitti's documentation contains two `StudentSetup.md` files, one under `docs/Developer` and the other under `docs/User`. Both files are linked in the `Table of Contents` and contain nearly the same information; the file under `docs/Developer` appears to be an updated version of the other one.

Oelgazz:
Fixes #510 by removing the file named "StudentSetup" in docs/User. The file under docs/User was removed instead of docs/Developer under the assumption that the majority of students using the simulator are contributors and under the fact that the file under docs/Developer is more updated.


stiber (changes requested):
Doesn't a link to this file also need to be either removed or redirected? Or are there no links to this file?

stiber (approved):

PR #615, opened 2024-03-10
[ISSUE 504] Add RecordableBase interface as the base class of matrix and add a RecordableVector template in Recorder

### LINKED ISSUE #504: Add an additional subclass/concrete class to the Recordable and make it working with the XmlRecorder
The objective of this issue is to add a new subclass/concrete class to the recordable framework so that the XmlRecorder can work with it effectively.

xiangli6:
closes #504 

1. Added Recorder::UpdateType to enable recording of both Dynamical and Constant variables
2. Added RecordableVector template to record standard library vector variant
3. Used RecordableBase interface as the base class of Matrix for recording VectorMatrix variables. Instead of creating a separate constantBuffer concrete class (initial plan), the RecordableBase interface was utilized as the base class of Matrix for recording VectorMatrix variables. This design decision was made to avoid disruption to existing code in the neural network simulation component.
4.  Did unit testing for XMLRecorder.
5. Updated variant to support uint64_t, bool, int, BGFLOAT types

Note: Although XmlRecorder now supports recording locations, this pull request does not include changes to output files.


Simulator/Recorders/Recordable.h
```diff
@@ -24,11 +24,11 @@ template <typename T> class Recordable : public RecordableBase {
    /// set up a string representing the basic data type
    virtual void setDataType() override
    {
-      basicDataType_ = "variant";
+      basicDataType_ = "T";
```
stiber, line 27: Hate to be annoying, but the reader will wonder: what does "T" mean?

Simulator/Vertices/EventBuffer.h
```diff
@@ -50,17 +50,12 @@ class EventBuffer : public Recordable<uint64_t> {
    /// Get the value of the recordable variable at the specified index.
    /// @param index The index of the recorded value to retrieve.
    /// @return A variant representing the recorded value (uint64_t, double, or string).
-   virtual variant<uint64_t, double, string> getElement(int index) const override;
+   virtual variantTypes getElement(int index) const override;
 
    /// Get A string representing the data type of the recordable variable
-   virtual const string &getDataType() const override;
+   virtual string &getDataType() override;
```
stiber, line 56: A question: this can't be `virtual string &getDataType() const override;`?

stiber (changes requested):
A couple questions.

Simulator/Recorders/Recordable.h
xiangli6: Yes, it's a bit unusual, and currently, I can't think of a better solution to improve it. Here's why this template returns a "T":

In the Recordable module, there's a data member called basicDataType_, which is a string representing the basic data type of the data stored in this class. However, for the template Recordable, since it doesn't know what type it will hold, it returns a "T". When users utilize it for variable registration without specifying the data type contained in this class, they need to manually pass the basic data type through the overloaded registerVariable function. This approach also applies to the template RecordableVector when registering standard library Vector variables in the variable owner class. Do you have any suggestions for improvement?

Simulator/Vertices/EventBuffer.h
xiangli6: Currently, the getDataType function is not marked as const because sometimes it changes the data type. I guess the correct approach to updating the value of basicDataType is to create a constructor in each concrete class and call setDataType() within each constructor of the concrete class. I mean, when an object of a concrete class is instantiated, its constructor should set the appropriate data type using setDataType(). This ensures that the basicDataType is updated correctly according to the type of object being created. Is my understanding correct? should I change code by adding constructors and const?

Simulator/Recorders/Recordable.h
stiber: I guess we would need to talk about this for me to understand the fine points of this. I was just getting at something simpler. This is just a string, so the "T" is literally what the string is; it doesn't get replaced with whatever the template is instantiated with. So, isn't this default then an "undefined" type? Perhaps the string should be "undefined"? That would be more-or-less self-documenting.

Actually, I see this line of code for `RecordableVector`. In that case, this really needs to be filled in with a string that will represent the actual type that the template `T` is bound to and your `typeid(T).name` seems to be the way to go.

Simulator/Vertices/EventBuffer.h
stiber: This is definitely something that is complex enough that we would need to talk. If you have the time, maybe we can go over the documentation and I can ask stupid questions.

xiangli6: Hello Professor, I am happy to have a meeting to go over this. I am available between 9 am and 2 pm, with the exception of Fridays when I am available until 1 pm. If these time slots are not convenient for you, please let me know, as my schedule is quite flexible and I can find an alternative time.

Simulator/Recorders/Recordable.h
xiangli6: I did a little more research about typeid(T). It indicates that using typeid can lead to inconsistencies across compilers. This is because the string returned by typeid(T).name() is implementation-dependent, meaning it can vary between different compilers and platforms. I believe we should avoid relying on the exact string returned by typeid(T).name() for cross-compiler compatibility.  Should I just use the string "undefined" as the basic data type in the Recordable and RecordableVector templates?

Simulator/Vertices/EventBuffer.h
stiber: From our conversation, this should actually be reverted to the old code, a const method returning a const reference to a strong.

Simulator/Recorders/XmlRecorder.cpp
```diff
@@ -137,13 +140,16 @@ string XmlRecorder::toXML(const string &name, vector<multipleTypes> &singleBuffe
...
          os << get<uint64_t>(element) << " ";
       } else if (basicType == "double") {
-         os << get<double>(element) << " ";
-      } else if (basicType == "string") {
-         os << get<string>(element) << " ";
+         os << get<bool>(element) << " ";
+      } else if (basicType == "int") {
+         os << get<int>(element) << " ";
+      } else if (basicType == "BGFLOAT") {
+         os << get<BGFLOAT>(element) << " ";
```
stiber, line 152: Add one more `else if`, to compare with `"undefined"` and generate an error message.

stiber (changes requested):

Simulator/Recorders/RecordableBase.h
```diff
@@ -14,35 +14,45 @@
...
 #include <vector>
 
+/// A list of pre-defined basic data types for variablse in all the simulations
+/// These pre-defined types should match with the types of variant in Recorder
+using variantTypes = variant<uint64_t, bool, int, BGFLOAT>;
+
 class RecordableBase {
-   /// TODO: will change function names later
 public:
    virtual ~RecordableBase() = default;
```
stiber, line 27: So, I did a little poking around. This class is fine as it is, in that if somehow someone wrote some code that instantiated a `RecordableBase` object, they will get a runtime error. However, if we want, we may be able to do one of the following to generate a compile-time error:

1. Declare the destructor as `virtual ~ RecordableBase() = 0;`, but then also provide the implementation outside of the class `RecordableBase::~ RecordableBase() {}`.
2. Declare a `protected` constructor.

The 1st option makes this an abstract base class; the latter prevents any code outside this class from creating a `RecordableBase` object. Supposedly.

KamathScience:
As far as I know, cereal serialization does not support classes inheriting directly from standard library containers. Since it's crucial to provide a `serialize` function for all base classes and we cannot add a `serialize` function to a standard library class.

To resolve this, I modified the `RecordableVector` class by removing its inheritance from `std::vector` and instead added it as a private member variable. This change ensures that serialization is compatible with cereal while allowing continued use of the vector functionality within `RecordableVector`.

@xiangli6 please make any adjustments as needed.

xiangli6:
> As far as I know, cereal serialization does not support classes inheriting directly from standard library containers. Since it's crucial to provide a `serialize` function for all base classes and we cannot add a `serialize` function to a standard library class.
> 
> To resolve this, I modified the `RecordableVector` class by removing its inheritance from `std::vector` and instead added it as a private member variable. This change ensures that serialization is compatible with cereal while allowing continued use of the vector functionality within `RecordableVector`.
> 
> @xiangli6 please make any adjustments as needed.

Here is the reason of the RecordableVector have multiple inheritance. Recorder received a list of different simulations variables with different types and structures and store these variable into a variable table. Therefore these recordable variables need a common base class RecordableBase. For vector<int> , it is hard to deal it without interrupting the existing code in variable owner class if not use recordableVector which is inherit from two RecordableBase and vector.

Simulator/Recorders/RecordableBase.h
xiangli6: I chose the second solution. The first solution generates an error when serializing this class

Simulator/Edges/Neuro/AllDSSynapses.h
```diff
@@ -49,7 +49,7 @@
 
 #include "AllSpikingSynapses.h"
 // cereal
-#include <cereal/archives/xml.hpp>   // this is a cereal hack
+// #include <cereal/archives/xml.hpp>   // this is a cereal hack
```
KamathScience, line 52: I am so sorry about this, but this cereal hack should not be commented. I forgot to uncomment this statement when pushing my change.

stiber: OK, adding this to the review.

Simulator/Recorders/Recordable.h
```diff
@@ -45,30 +55,21 @@ template <typename T> class Recordable : public RecordableBase {
...
-      return basicDataType_;
-   }
-
    ///  Cereal serialization method
    template <class Archive> void serialize(Archive &archive)
    {
       archive(cereal::virtual_base_class<RecordableBase>(this),
-              cereal::make_nvp("eventTimeSteps_", eventTimeSteps_),
-              cereal::make_nvp("basicDataType_", basicDataType_));
+              cereal::make_nvp("eventTimeSteps_", eventTimeSteps_));
```
stiber, line 67: I think this variable name needs to be changed to something more generic than "event...".

Simulator/Recorders/RecordableVector.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include "RecordableBase.h"
+#include <stdexcept>   // for std::out_of_range
+#include <vector>
+// Cereal
+#include <cereal/types/polymorphic.hpp>
+#include <cereal/types/vector.hpp>
+
+template <typename T> class RecordableVector : public RecordableBase, public std::vector<T> {
+public:
```
stiber, line 11: So, @KamathScience pointed out an issue with inheriting from `vector`. Is that resolved?

stiber (changes requested):
Mostly questions that we just need to confirm are dealt with.

Simulator/Recorders/RecordableVector.h
xiangli6: @KamathScience Would you give a check to the serialization code in the RecordableVector class? Thanks.

Simulator/Recorders/Recordable.h
xiangli6: renamed it from eventTimeSteps_ to dataSeries_ in the RecordableVector class.

Simulator/Recorders/RecordableVector.h
KamathScience: I tested the simulation on both CPU and GPU for serialization (test-small-connected and test-medium-connected) and these recent modifications work like a charm! Thank you @xiangli6

stiber (commented):
Mostly small stuff, and a question.

Simulator/Vertices/EventBuffer.cpp
```diff
@@ -77,31 +65,37 @@ void EventBuffer::clear()
...
 uint64_t EventBuffer::operator[](int i) const
 {
-   return eventTimeSteps_[(epochStart_ + i) % eventTimeSteps_.size()];
+   return dataSeries_[(epochStart_ + i) % dataSeries_.size()];
 }
 
+void EventBuffer::startNewEpoch()
+{
+   epochStart_ = queueEnd_;
+   queueFront_ = queueEnd_;
```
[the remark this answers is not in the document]
KamathScience: Yes, I will work on that in a different issue

Simulator/Vertices/EventBuffer.cpp
```diff
@@ -42,27 +41,16 @@ const string &EventBuffer::getDataType() const
...
-
 int EventBuffer::getNumElements() const
-{
-   return getNumEventsInEpoch();
-}
-
-int EventBuffer::getNumEventsInEpoch() const
 {
    return numEventsInEpoch_;
 }
```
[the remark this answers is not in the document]
xiangli6: I see. Indeed there is a little bit distinction here. I will fix it.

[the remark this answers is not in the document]
xiangli6: This is a bit of confusing for me too.
 In the RecordableVector class, the number of element in the vector is the number of elements in the current epoch but in the EventBuffer class, the number of elements in the vector not equals to the number of elements in the current epoch and the Recorder only wants to get the number of elements in each epoch from the EventBuffer. This is the reason why I decide to merge it and overlook the distinction. We can have two virtual methods size() and getNumElementInEpoch() in both RecordableVector and EventBuffer. Does this make sense?

Simulator/Vertices/EventBuffer.cpp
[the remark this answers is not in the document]
xiangli6: I can fix it here.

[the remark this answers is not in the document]
KamathScience: Sure, the GPU counterpart might also needs to be changed accordingly.

Simulator/Vertices/EventBuffer.cpp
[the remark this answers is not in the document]
stiber: OK, thinking further. A `RecordableVector` is just an ordinary-appearing vector, but with the extras needed to make it `Recordable` (regardless of what its parent class actually is). Any class that uses it manipulates it like a `vector` with those additional `Recordable` methods (the care simulator doesn't use those extra methods, of course, so it just is used like a `vector` there).

On the other hand, an `EventBuffer` has a second interface: to the `Recorder` classes, it looks like a `RecordableVector`, but to the core simulator, it looks like a buffer: it can be written to and past elements can be retrieved. So, the current `getNumEventsInEpoch()` method is the `vector` size operation. But, there is a difference with `RecordableVector`, in that after the `Recorder` is done with it and "clears" it, the buffer state isn't modified — it looks like it's empty when you think of it as a `vector`, but nothing has changed when you think of it as a buffer. So, there is a need to maintain the buffer state information, which means we need to know where the most recent event is and we need to know how many events are in the buffer. That's because, until the buffer fills up, wraps around, and new events start over-writing the oldest events, the number of events in the buffer is smaller than the buffer capacity. Of course, once it wraps around, the number of events stops increasing.

So, maybe we don't need a method to return the number of elements in the buffer; we just need a member variable.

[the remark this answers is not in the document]
xiangli6: The Recorder needs to get the number of elements in the RecordableVector object(the size of vector) while it needs to get the number of event in the current epoch from the EventBuffer. If we only have one such method in common interface, the RecordableBase, maybe we can document them in the comment. For example
RecordableBase{
    // get the number of element that need to be record in the object
    virtual int getNumElements() constant;
} 
RecordableVector: public RecordableBase{
    virtual int getNumElement() constant override
}
EventBuffer: public RecordableVector{
    //we need this virtual method because it is called by the Recorder
    virtual int getNumElement() constant override{
       return getNumElementInEpoch or numEventsInEpoch;
    }
    //we still need this method because it is called by variable owner class such as AllSpikingNeurons
    int getNumElementInEpoch() constant{
          return numEventsInEpoch
    }
private: 
    numEventsInEpoch_// there already is a data member showing showing the events information, maybe I need to rename it as numElementsInEpoch.
}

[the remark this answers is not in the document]
xiangli6: I will rename the data member numEventsInEpoch_ as numElementsInEpoch_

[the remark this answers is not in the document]
stiber: Does the variable owner class need to call `EventBuffer:: getNumElementInEpoch()`, or is that just needed within the `EventBuffer` class? If the latter, then we don't need that accessor; we can just use the data member.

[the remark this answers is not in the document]
xiangli6: The variable owner class, AllSpikingNeurons, needs to call this method.
```c++
//data member
vector<EventBuffer> vertexEvent_
// in cpp file, 
 vertexEvent[I].getNumElementInEpoch()) 
```
So I think we still need to keep it.

PR #623, opened 2024-03-28
[ISSUE-606] Remove Additional Slot from Event Buffer 

### LINKED ISSUE #606: Assertion failure in HDF5Recorder class during large GPU simulation
The simulator encounters an assertion failure within the `compileHistories()` method of the `Hdf5GrowthRecorder` class when executing a full-size growth simulation on the GPU at epoch 33.

 This issue is seen in `version 0.9.1` of the Graphitti, whereas `version 0.9` does not display this problem.

image

KamathScience:
Closes #606 

This PR addresses a bug related to the event buffer implementation. The issue stemmed from an additional slot in the event buffer, leading to inconsistencies between CPU and GPU implementations. This inconsistency caused unexpected behavior during GPU simulation.

Changes Made:
Removed the additional slot from the event buffer initialization, ensuring consistency between CPU and GPU implementations and enhanced the condition for detecting a full buffer using `numEventsInEpoch_`.
Adjusted the buffer size on the GPU side to match the modified event buffer, ensuring compatibility with the CPU implementation.

- [x] Tested for GPU with 12 configuration files. GPU testing was also conducted on `test-large-very-long.xml` file for both `XmlRecorder` and `Hdf5GrowthRecorder` till epoch 50.

### DIFF
--- a/Simulator/Vertices/EventBuffer.cpp
+++ b/Simulator/Vertices/EventBuffer.cpp
@@ -13,14 +13,9 @@
 #include <cassert>
 #include <limits>
 
-// EventBuffer::EventBuffer(int maxEvents) :
-//    eventTimeSteps_(maxEvents + 1, numeric_limits<unsigned long>::max())
-// {
-//    clear();
-// }
 EventBuffer::EventBuffer(int maxEvents)
 {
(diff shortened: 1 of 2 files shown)


stiber (approved):
Assuming that our recent conversation doesn't modify anything here, this is good to go.

Issue #634, opened 2024-04-29
Modify the `GraphManager` class so unit tests can be written without needing a `ParameterManager` object


jardiamj:
Currently, the `GraphManager` class needs a `ParameterManager` object to get the path to the `graphML` file from a configuration file. It contains a method to set the path to the `graphML` file however, that is overwritten in the `readGraph` method. In the `readGraph` method we need to check if that path has already been set and only retrieve it from the `ParameterManager` if it hasn't.
This will allow us to write Unit Tests for the `GraphManager` independently from the `ParameterManager` class and without configuration files.

jardiamj:
Fixed by #635

PR #636, opened 2024-05-02
[ISSUE 632] Graph Manager Unit Tests

### LINKED ISSUE #632: Create Unit Tests for Graph Manager
Since we are expanding Graph Manager's use (currently being integrated into neural network simulations), we'll need to create unit tests to ensure that Graph Manager is working properly.

Oelgazz:
closes #632


Oelgazz:
Currently 4 of the Graph Manager's tests fail. This is because of the Parameter Manager's dependency that is outlined in issue #634

jardiamj:
We indent to 3 spaces in our code base. I recommend adjusting the indentation spaces in your IDE or our clang-format check will fail.

jardiamj:
I know that most of the methods in the `GraphManager` class are thin wrappers around the Boost Graph Library but it might be a good idea to write tests for them. One could still make mistakes and break functionality. These tests would serve as a check against breaking functionality and also demonstrate how the `GraphManager` works.

jardiamj:
The `GraphManagerTests.cpp` file doesn't pass the clang-format check because it uses the Windows-style end-of-line characters (`\r\n`) at the end of each line, instead of the Linux `\n` one. If you open the file with `vim -b` you will see a `^M` character at the end of each line, which is the way `\r` is rendered.

You must convert that file to the Linux end-of-line character (`\n`).

jardiamj (approved):

PR #638, opened 2024-05-08
[ISSUE 270] Handle Nullptr From Factory

### LINKED ISSUE #270: Asking the Factory to create a nonexistent class should generate a FATAL error
When the name of the class defined in the config files is wrong, the factories return nullptr, which leads to a seg fault. This should instead be a FATAL error.
This also happens when the name string is uninitialized. Variables that couldn't be initialized by `ParameterManager` should also throw FATAL errors (see issue #217 , which, given our now longer regression test times, should probably be a higher priority).

Oelgazz:
Closes #270 
Since Google Test acts up when we test for an exit. We will keep factory returning nullptr, but each class that uses the factory will check for nullptr, and will exit the program if that is the case.

### DIFF
--- a/Simulator/Connections/Connections.cpp
+++ b/Simulator/Connections/Connections.cpp
@@ -24,19 +24,10 @@
 
 Connections::Connections()
 {
-   // Get a copy of the console logger to use in the case of errors
-   log4cplus::Logger consoleLogger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
-   
    // Create Edges/Synapses class using type definition in configuration file
    string type;
    ParameterManager::getInstance().getStringByXpath("//EdgesParams/@class", type);
    edges_ = Factory<AllEdges>::getInstance().createType(type);
-
-   // If the factory returns an error (nullptr), exit
-   if (edges_ == nullptr) {
-      LOG4CPLUS_INFO(consoleLogger, "INVALID CLASS: " + type);
-      exit(EXIT_FAILURE);
-   }
 
    // Get pointer to operations manager Singleton
    OperationManager &opsManager = OperationManager::getInstance();
--- a/Simulator/Connections/Connections.cpp
+++ b/Simulator/Connections/Connections.cpp
@@ -33,7 +33,7 @@ Connections::Connections()
    edges_ = Factory<AllEdges>::getInstance().createType(type);
 
    // If the factory returns an error (nullptr), exit
-   if (edges_ == nullptr) {
+   if (!edges_) {
       LOG4CPLUS_INFO(consoleLogger_, "INVALID CLASS: " + type);
       exit(EXIT_FAILURE);
    }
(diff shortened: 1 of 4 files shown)


Simulator/Connections/Connections.cpp
```diff
@@ -24,11 +24,20 @@
...
+   // Get a copy of the console logger to use in the case of errors
+   log4cplus::Logger consoleLogger_ = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
+   
    // Create Edges/Synapses class using type definition in configuration file
    string type;
    ParameterManager::getInstance().getStringByXpath("//EdgesParams/@class", type);
    edges_ = Factory<AllEdges>::getInstance().createType(type);
 
+   // If the factory returns an error (nullptr), exit
+   if (!edges_) {
```
stiber, line 36: This uses an implicit/automatic conversion of a pointer type to a `bool`. It's better, and clearer, to use `if (edges_ == nullptr)`

Simulator/Core/Simulator.cpp
```diff
@@ -89,11 +89,20 @@ void Simulator::loadParameters()
       deltaT_ = DEFAULT_dt;
    }
 
+   // Note: This might be unnecessary, but to ensure that consoleLogger is initialized
+   consoleLogger_ = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
+
```
stiber, line 94: `Simulator` already has a `consoleLogger_` member variable, which is assigned to in the constructor. So, it is safe to use.

stiber (changes requested):
Some changes. Note that the comment for `Connections` applies to some of the other classes (but not all).

Simulator/Connections/Connections.cpp
```diff
@@ -24,11 +24,20 @@
 
 Connections::Connections()
 {
+   // Get a copy of the console logger to use in the case of errors
+   log4cplus::Logger consoleLogger_ = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
```
stiber, line 28: If you want this to be a local variable, it should not have the trailing "`_`".

Simulator/Core/Model.cpp
```diff
@@ -23,21 +23,42 @@
 /// Constructor
 Model::Model()
 {
+   // Get a copy of the console logger to use in the case of errors
+   log4cplus::Logger consoleLogger_ = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
```
stiber, line 27: If you want this to be a local variable, it should not have the trailing "_".

Simulator/Layouts/Layout.cpp
```diff
@@ -18,11 +18,20 @@
 /// Constructor
 Layout::Layout() : numEndogenouslyActiveNeurons_(0)
 {
+   // Get a copy of the console logger to use in the case of errors
+   log4cplus::Logger consoleLogger_ = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("console"));
```
stiber, line 22: If you want this to be a local variable, it should not have the trailing "_".

stiber (changes requested):
I think there's a good chance that fixing these things will allow the automated tests to pass.

stiber (approved):

PR #649, opened 2024-05-20
[ISSUE 642] Use GraphML For Endogenously Active Neurons

### LINKED ISSUE #642: Set which neurons are endogenously active from GraphML file
This can be a boolean attribute for each vertex.

Oelgazz:
Closes #642


KamathScience (commented):
Great work Omar! Serialization changes look good to me.

Tools/NeuroGraphGenerator.py
```diff
@@ -0,0 +1,38 @@
...
+# Credit to Jardi A.M. Jordan for initial code
+
+# IMPORTANT: Once you generate the GraphML file, you will need
+# to manually change the keys "attr.type" member to the data type you desire
+# For example for key 'x' you will need to change attr.type from "long" to "double"
+# and for key 'active' you will need to change attr.type from "long" to "boolean"
+
+# It is currently unknown whether the networkx library has a way to change the data
+# type automatically, but from the research conducted this can't be done in this script
+# and thus must be done manually.
```
jardiamj, line 16: You can define the data type in this script by ensuring the values stored are of the desired type. For instance, to get active to be a boolean we can do `node['active'] = True`.

In the case of doubles, we would cast the values to a float (floats in Python are the equivalent of doubles in C++). For example:
```
node['x'] = float(id % height)
```

jardiamj (changes requested):
I am being picky with this review.

Tools/NeuroGraphGenerator.py
Oelgazz: 'x' and 'y' have been updated. For 'active', the reason why I can't use "True" is that the GraphML is generated with the word "True" in the key. When I run the boost, it gives me this error:
terminate called after throwing an instance of 

'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::parse_error> >'
  what():  parse error: invalid value "True" for key active of type boolean
Aborted

note that this is case insensitive. The only way to have a boolean is to either use a number and change the key's `attr.type` from "long" to "boolean", or to change this script from python to C++

jardiamj (approved):

stiber (approved):

PR #679, opened 2024-07-22
Implement the initDataSet, actually is 'saveSimData' for updated type CONSTANT

### LINKED ISSUE #657: Implement Hdf5Recorder::initDataset() for UpdatedType::CONSTANT
The goal of this is to demonstrate that we can create an HDF5 file with data in it that is set at the start of the simulation and doesn't change.

Hsu0227:
Closes #657 

#### Description
This time I implemented the saveSimData function to replace Hdf5Recorder::initDataset() for UpdatedType::CONSTANT, which is responsible for saving simulation data into HDF5 datasets. I decide to use two different function for different UpdatedType to save and store data. Because after I see the XML recorder, I think saveSimData is more matches the initDataset() for UpdatedType::CONSTANT, I think for our new hdf5 recorder we don't have to have the initDataset(), and also for the latter task initDataset() for UpdatedType:: DYNAMIC we can use compileHistories() to responsible instead, if we really need to use initDataset() we can both find some way to call compileHistories() and saveSimData()  in initDataset() after I implementer both function. And also this time create a helper function- captureData, to ensure that the dataset exists and that data can be written to it.


Simulator/Recorders/Hdf5Recorder.h
```diff
@@ -127,31 +129,49 @@ class Hdf5Recorder : public Recorder {
          if (dataType_ == typeid(uint64_t).name()) {
             hdf5Datatype_ = PredType::NATIVE_UINT64;
          } else if (dataType_ == typeid(bool).name()) {
-            hdf5Datatype_ = PredType::NATIVE_UINT8;
+            hdf5Datatype_ = PredType::NATIVE_INT;
          } else if (dataType_ == typeid(int).name()) {
             hdf5Datatype_ = PredType::NATIVE_INT;
-         } else if (dataType_ == typeid(float).name()) {
+         } else if (dataType_ == typeid(BGFLOAT).name()) {
```
stiber, line 135: Actually, I think the original here (comparing to `float`) is better. `BGFLOAT` will either be a `float` or a `double`, and this will then work appropriately either way.

Simulator/Recorders/Hdf5Recorder.h
```diff
@@ -127,31 +129,49 @@ class Hdf5Recorder : public Recorder {
...
             hdf5Datatype_ = PredType::NATIVE_FLOAT;
-         } else if (dataType_ == typeid(double).name()) {
-            hdf5Datatype_ = PredType::NATIVE_DOUBLE;
```
stiber, line 136: Why delete this? Would it cause a problem?

Simulator/Recorders/Hdf5Recorder.h
Hsu0227: I will change it back

stiber (approved):

PR #690, opened 2024-08-06
Implement Hdf5Recorder::registerVariable() for vector of RecordableBase

### LINKED ISSUE #659: Implement Hdf5Recorder::registerVariable() for vector of RecordableBase

Hsu0227:
Closes #659 

#### Description
I Implement Hdf5Recorder::registerVariable() for vector of RecordableBase, the code inside is the same, but I can't remove it because for the singleVariableInfo , the code inside is different, and inside the registerVariable we gonna use this class. And I also remove some functions that we don't need like initdefaultvalue, initvalue, and getvalue. And after that I updated the documentation.


Hsu0227:
I think this is my last pull request, right? I remember the last time we talked about I don't need to Refactor the code since I don't have enough time.

stiber (approved):
Looks good!

PR #733, opened 2024-11-19
[ISSUE-680] Record Zero Length Variables

### LINKED ISSUE #680: Recorders should output zero-length variables
`XMLRecorder::saveSimData()` shouldn't do a check of the number of elements in a variable before outputting it.

varndorfer:
Closes #680 

#### Description
XML Recorder should produce an empty output for variables with no values. Tested and now the output is as in the below image: 
image

- [ ] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/Simulator/Recorders/XmlRecorder.cpp
+++ b/Simulator/Recorders/XmlRecorder.cpp
@@ -83,9 +83,9 @@ void XmlRecorder::saveSimData()
       }
       // cout << variableTable_[rowIndex].variableName_ << endl;
       resultOut_ << toXML(variableTable_[rowIndex].variableName_,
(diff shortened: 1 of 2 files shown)


stiber:
Are any changes needed to `Hdf5Recorder`?

stiber:
So, do we really want to do this? Let's discuss it at a lab meeting; I think there are arguments either way, but maybe not; maybe this was just something different than before. After all, if we don't output a variable, then we know 100% what's in it. OTOH, if we register a variable and it doesn't appear in the output, we can't tell if that means it was empty or if something is broken in the simulator. Hmm...

PR #767, opened 2025-01-17
[ISSUE-753] CMakeLists fixes

### LINKED ISSUE #753: CMakeLists needs to have its CUDA files/library creation cleaned up
Right now, it looks like all of the CUDA files get added to a STATIC library and to a bunch of SHARED libraries. They are also individually deleted from libraries (but maybe not all; maybe just the ones under each subdirectory) explicitly. Instead, we might make all libraries STATIC and only include the relevant CUDA files in each library. Or, alternatively, we might make a single STATIC library with all CUDA files, putting none in with the various subsystems.

This needs to be figure out by someone who really understands CMake.

https://github.com/UWB-Biocomputing/Graphitti/blob/7a50b56ad048c1bdb610d5b6e1351a3e1640ea1a/CMakeLists.txt#L162-L191

AndrewBMadison:
Closes #753

#### Description
I was unable to run ggraphitti with the old CMakeLists configuration. I was able to compile and link, but when the program loaded there was a seg fault. I investigated and there is an external variable, classSynapses_d, that is defined in AllSynapsesDeviceFuncs_d.cpp. Under the old CMakeLists configuration, this file was included in multiple libraries, two of which were linked together. I am not sure how this works.

I changed the source files for the three libraries, edges, vertices, and connections, to only include the necessary sources for each library.  In addition to this change, I made the libraries static. This solved the problem I had, and now ggraphitti loads and runs correctly. 

- [ ] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,8 +46,8 @@ if(ENABLE_CUDA)
         add_compile_definitions(USE_GPU)
 #Specify the CUDA architecture / gencode that will be targeted
         ### Set gencode and architecture variables to the correct values for your specific NVIDIA hardware
-        set(CMAKE_CUDA_ARCHITECTURES 75)        
-        set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode=arch=compute_75,code=sm_75)
+        set(CMAKE_CUDA_ARCHITECTURES 37)        
+        set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode=arch=compute_37,code=sm_37)
             
 else()
         message("\n----Generating Makefile for Graphitti CPU version----")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ cmake_minimum_required(VERSION 3.12)
 #
 #"YES" / GPU choice only available if CUDA library is installed and the GPU is CUDA capable.
 ############################################################################################
+
 if(NOT ENABLE_CUDA)
         set(ENABLE_CUDA NO)
 endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -45,8 +46,8 @@ if(ENABLE_CUDA)
         add_compile_definitions(USE_GPU)
 #Specify the CUDA architecture / gencode that will be targeted
         ### Set gencode and architecture variables to the correct values for your specific NVIDIA hardware
-        set(CMAKE_CUDA_ARCHITECTURES 37)        
-        set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode=arch=compute_37,code=sm_37)
+        set(CMAKE_CUDA_ARCHITECTURES 75)        
+        set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode=arch=compute_75,code=sm_75)
```
stiber, line 50: I think we need to consider what the implications are of changing these compilation parameters. I believe that there is an issue that asks someone to investigate this. Besides understanding and documenting the architecture models' impact on Graphitti, we would need to evaluate its effects on results (regression testing) and performance.

stiber (changes requested):
Leaving out the architecture flag matter, the changes look good. But I'm not clear about the variable indicated in the PR description, because that doesn't look like a variable and doesn't seem to be in the changes made. In addition, the fact that you weren't able to compile and run is a bit concerning, since we know that we are able to compile and run the simulator. Might be good to consult with @NicolasJPosey .

CMakeLists.txt
AndrewBMadison: I changed the architecture back to kepler.

I am currently running CUDA toolkit version 12.6. which does not support the kepler architecture. I will update my local CMakeLists.txt to a supported architecture when I compile and remember to change it back before I push it.

As a side note, I found this paper which will help us when we evaluate CUDA architectures and their performance implications: https://www.sciencedirect.com/science/article/pii/S016781912400019X

stiber: OK. Remember that the GPU version needs to be tested on raiju/otachi. In particular, if there are problems compiling on your machine but not on our research machines, that is a matter that needs to be resolved by determining the root cause.

AndrewBMadison:
I wanted to provide a complete update regarding the segmentation fault issue in GGraphitti. After investigation, I discovered that two shared libraries were each including AllSynapsesDeviceFunc.cpp (and its corresponding header). This file declares a device variable

`extern __device__ enumClassSynapses classSynapses_d;`

Because CUDA does not support having the same global device variable defined across multiple shared objects, the loader could not reconcile these duplicate definitions at runtime, resulting in a segmentation fault.

To confirm this, I used nm to inspect each library and saw that classSynapses_d appeared in both. This confirmed the collision. Unlike CPU linkers, CUDA cannot unify duplicated device symbols at load time, leading to the undefined behavior.

Resolution:
I removed the duplicate AllSynapsesDeviceFunc.cpp source file from all but one library so that classSynapses_d is defined only once. I then switched to static linking for the affected libraries, ensuring the device variable is consolidated into a single compilation unit. After these changes, the application now runs correctly without segmentation faults.
Please let me know if you have any questions or would like more information on the steps taken.

stiber (approved):

PR #777, opened 2025-01-26
[issue-745] Refactor where summation is calculated

### LINKED ISSUE #745: Refactor summation point logic in CPU neuron model to better align with GPU model
The GPUModel.cpp has a calcSummationPoint method that implements the GPU version of the summation point logic. On the CPU side, this logic is embedded in the AllSpikingSynapses::advanceEdge method.

This should be pulled out into a calcSummationPoint method that the CPUModel.cpp implements. This method would then get called in the CPUModel::advance method after advancing the edges, similar to to how the GPUModel::advance method behaves.

Some other related changes that could be made:

- Since CPUModel and GPUModel would both be implementing a calcSummationPoint method, this could be added as a virtual method in Model.h for a derived type of model to implement.
- The name "calcSummationPoint" would not be an appropriate name for an NG911 model. Therefore, if we make change 1, we should rename "calcSummationPoint". I propose this method is named "updateVertexState".
- In addition, we wouldn't expect the NG911 model to have the same implementation logic as a neuron model for updateVertexState. I see two ways to resolve this issue:
  1. The first is that we could consider changing the current "CPUModel" and "GPUModel" classes to "NeuronCPUModel" and "NeuronGPUModel" respectively. This would allow us to define different logic for updating the vertex state in a NG911 model while keeping the changes relatively minimal. I estimate that at least the Simulator.cpp class and the Model classes would need to be changed (assuming we can get the derived type information from what is currently in the config files).
  2. The second is that we could add a updateVertexState method to the AllVertices class and call updateVertexState as a AllVertices method instead of a Model method. This would allow vertex-related stuff to live in a more appropriate place.

The last point can be summarized as such:

- Make Model be extensible for neuron and NG911 models
  - Pros: Implementation would mostly look like it currently does for the area we are trying to update.
  - Cons: Model type would contain vertex-related logic. Vertices are already extensible for neuron and NG911 models. Introduces new model types which may require changes to config files.
- Add current calcSummationPoint logic into AllVertices
  - Pros: Cleaner separation of model and vertex logic. No need to add new derived model types.
  - Cons: Less files being touched, but are touched more (e.g. the summation point kernel would need to be moved from the GPUModel.cpp into the AllVertices_d.cpp).

NicolasJPosey:
Closes #745 

#### Description
The summation point logic was being done in the neuron edges for the CPU model while the GPU model had a calcSummationPoint method that handled the GPU-side summation point logic. The feature branch refactors this by introducing a integrateVertexInputs method to AllVertices for both CPU and GPU. The neuron implementation of AllVertices can then implement the summation point logic in this method. This refactor also allows models types other than neuron (such as 911) to define how they want to integration their vertex inputs since not all model types will use summation point logic.

- [ ] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/Simulator/Core/GPUModel.cpp
+++ b/Simulator/Core/GPUModel.cpp
@@ -153,9 +153,8 @@ void GPUModel::advance()
 
    // display running info to console
    // Advance neurons ------------->
-   dynamic_cast<AllSpikingNeurons &>(vertices).advanceVertices(edges,
-                                                              allVerticesDevice_, allEdgesDevice_,
-                                                              randNoise_d, synapseIndexMapDevice_);
+   dynamic_cast<AllSpikingNeurons &>(vertices).advanceVertices(
+      edges, allVerticesDevice_, allEdgesDevice_, randNoise_d, synapseIndexMapDevice_);
 
 #ifdef PERFORMANCE_METRICS
    cudaLapTime(t_gpu_advanceNeurons);
@@ -171,7 +170,8 @@ void GPUModel::advance()
 #endif   // PERFORMANCE_METRICS
 
    // integrate the inputs of the vertices
-   dynamic_cast<AllSpikingNeurons &>(vertices).integrateVertexInputs(allVerticesDevice_, synapseIndexMapDevice_, allEdgesDevice_);
+   dynamic_cast<AllSpikingNeurons &>(vertices).integrateVertexInputs(
+      allVerticesDevice_, synapseIndexMapDevice_, allEdgesDevice_);
 
 #ifdef PERFORMANCE_METRICS
    cudaLapTime(t_gpu_calcSummation);
--- a/Simulator/Core/GPUModel.cpp
+++ b/Simulator/Core/GPUModel.cpp
@@ -153,7 +153,7 @@ void GPUModel::advance()
 
    // display running info to console
    // Advance neurons ------------->
-   dynamic_cast<AllSpikingNeurons &>(vertices).advanceVertices(
+   vertices.advanceVertices(
       edges, allVerticesDevice_, allEdgesDevice_, randNoise_d, synapseIndexMapDevice_);
 
 #ifdef PERFORMANCE_METRICS
@@ -170,7 +170,7 @@ void GPUModel::advance()
 #endif   // PERFORMANCE_METRICS
 
    // integrate the inputs of the vertices
-   dynamic_cast<AllSpikingNeurons &>(vertices).integrateVertexInputs(
+   vertices.integrateVertexInputs(
       allVerticesDevice_, synapseIndexMapDevice_, allEdgesDevice_);
 
 #ifdef PERFORMANCE_METRICS
--- a/Simulator/Core/GPUModel.cpp
+++ b/Simulator/Core/GPUModel.cpp
@@ -153,8 +153,8 @@ void GPUModel::advance()
 
    // display running info to console
    // Advance neurons ------------->
-   vertices.advanceVertices(
-      edges, allVerticesDevice_, allEdgesDevice_, randNoise_d, synapseIndexMapDevice_);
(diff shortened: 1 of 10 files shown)


NicolasJPosey:
@stiber ran the GPU unit and regression tests as well and they all passed so I think this change is ready for a review.

stiber:
Nick, did you want to request a review from me? That's done under "reviewers", not "assignees"...

NicolasJPosey:
@stiber yes, that was my intention. Sorry about that. I will switch that around.

Simulator/Vertices/NG911/All911Vertices.h
```diff
@@ -232,6 +232,15 @@ class All911Vertices : public AllVertices {
...
    virtual void setAdvanceVerticesDeviceParams(AllEdges &edges) {};
+
+   /// Performs an integration operation per vertex using the inputs to the vertex.
+   ///
+   /// @param allVerticesDevice       GPU address of the allVertices struct on device memory.
+   /// @param edgeIndexMapDevice      GPU address of the EdgeIndexMap on device memory.
+   /// @param allEdgesDevice          GPU address of the allEdges struct on device memory.
+   virtual void integrateVertexInputs(void *allVerticesDevice,
+                                      EdgeIndexMapDevice *edgeIndexMapDevice,
+                                      void *allEdgesDevice) {};
```
stiber, line 243: Implementations of inherited methods, even pure virtual, should be marked `override`.

stiber (changes requested):
Some small-ish things.

Simulator/Core/GPUModel.cpp
```diff
@@ -153,44 +153,31 @@ void GPUModel::advance()
 
    // display running info to console
    // Advance neurons ------------->
-   dynamic_cast<AllSpikingNeurons &>(neurons).advanceVertices(connections_->getEdges(),
-                                                              allVerticesDevice_, allEdgesDevice_,
-                                                              randNoise_d, synapseIndexMapDevice_);
+   dynamic_cast<AllSpikingNeurons &>(vertices).advanceVertices(
+      edges, allVerticesDevice_, allEdgesDevice_, randNoise_d, synapseIndexMapDevice_);
```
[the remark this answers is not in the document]
NicolasJPosey: Removed dynamic cast and nothing broke. GPU regression and unit tests are still passing. Running CPU regression tests one more time and then will commit and push my changes to this remote branch.

Simulator/Core/GPUModel.cpp
```diff
@@ -153,44 +153,31 @@ void GPUModel::advance()
...
 #ifdef PERFORMANCE_METRICS
    cudaLapTime(t_gpu_advanceSynapses);
    cudaStartTimer();
 #endif   // PERFORMANCE_METRICS
 
-   // calculate summation point
-   calcSummationPoint();
+   // integrate the inputs of the vertices
+   dynamic_cast<AllSpikingNeurons &>(vertices).integrateVertexInputs(
+      allVerticesDevice_, synapseIndexMapDevice_, allEdgesDevice_);
```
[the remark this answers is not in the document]
NicolasJPosey: Changes and GPU regression is still passing.

Simulator/Vertices/NG911/All911Vertices.h
NicolasJPosey: Done. Running CPU unit tests to confirm still passing.

NicolasJPosey: Found out that the 911 test config file doesn't get run with our CPU regression test. Put in an issue #786 to discuss updating RunTests.sh so that it is run with the other regression tests. Manually added it to RunTests.sh and confirmed that it's passing.

NicolasJPosey:
@stiber all changes have been made and all tests have been run and are passing.

Issue #785, opened 2025-02-01
Fix some remaining neuro specific items in GPUModel.h


stiber:
Such as:

https://github.com/UWB-Biocomputing/Graphitti/blob/291506208498bfaa2d5d13372c069dabe290c86c/Simulator/Core/GPUModel.h#L126-L130

NicolasJPosey:
Completed and pushed into SharedDevelopment

PR #798, opened 2025-02-18
Merge PoseyDevelopment into SharedDevelopment

### LINKED ISSUE #745: Refactor summation point logic in CPU neuron model to better align with GPU model

NicolasJPosey:
Closes #745 #785 

#### Description
Merge includes two PRs. PR 777 which moved the vertex integration logic that lived in the Edges class into the Vertices class and PR 789 which cleaned up neuro-specific references in the GPU and CPU models.

- [x] Added Documentation 
- [ ] Added Unit Tests 


NicolasJPosey:
@stiber this PR is to merge my PoseyDevelopment branch into SharedDevelopment. This will merge PRs 777 and 789 into SharedDevelopment. Per the above commits, it is now up-to-date with SharedDevelopment. I took a quick look through the file changes and it looks like it only contains the changes from the two PRs so should be a quick review. I ran both the CPU and GPU tests which all passed.

stiber:
OK, one thing we learn from this exercise is that git/GitHub doesn't track previously approved changes, so everything shows up here under "files changed". I think, given all of the changes have bee previously reviewed, the thing to examine here is that:

-  `SharedDevelopment` has been merged into the personal dev branch, if needed (I think you did that but I don't think that involved any changes, because your dev branch wasn't behind.
- Look at any post-merge changes needed to fix things that broke as a result of that merge (which isn't applicable here).

So, unless you can think of something else, I think I can approve this without detailed examination.

NicolasJPosey:
@stiber I can't think of anything beyond your two points. I did a sanity check and confirmed that for at least one of the files that I changed, a change made to that same file that was in SharedDevelopment and not my branch was successfully merged in. So I think because there was no merge conflicts that needed to be resolved, the file changes post-merge are just the new changes that I am introducing to SharedDevelopment.

PR #799, opened 2025-02-18
[ISSUE-648] Templatize GraphManager Class and Add Inheritance to VertexProperty

### LINKED ISSUE #648: Modify GraphManager so that it can handle different property structs for different application domains
Right now, we use a single struct for each kind of property (vertex, edge, graph), with all properties for all domains merged together. It would be good to instead either use inheritance in the structs or a template `GraphManager` to enable each domain to define its properties separately.

jasleenksaini:
Closes #648

#### Description
The GraphManager class needs to be modified to support different property structs for specific application domains. This enhancement will involve templating the class and adding inheritance to the VertexProperty.

- [x] Added Documentation 
- [x] Added Unit Tests 

### DIFF
--- a/Simulator/Layouts/Neuro/LayoutNeuro.cpp
+++ b/Simulator/Layouts/Neuro/LayoutNeuro.cpp
@@ -6,123 +6,123 @@
  * @brief The Layout class defines the layout of vertices in neural networks
  */
 
- #include "LayoutNeuro.h"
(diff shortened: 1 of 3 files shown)


stiber (changes requested):
Mostly just making that preceding changes in SharedDevelopment aren't reverted.

PR #804, opened 2025-02-28
[issue-786] Add 911 test to RunTests.sh

### LINKED ISSUE #786: 911 regression config not being run
Below is the output from a branch based off SharedDevelopment

Image

There is a "test-small-911.xml" in Testing/RegressionTesting/configFiles. It doesn't look like this is being run when we run the CPU regression tests.

I suspect it's because it needs to manually be added to the TEST_FILES in RunTests.sh

Image

@stiber is there a reason NOT to add the test-small-911.xml to TEST_FILES so that it is run with the other CPU tests? We might have to do an if else so that we only add it when the processing unit is CPU.

NicolasJPosey:
Closes #786

#### Description
It looks like the 911 file wasn't moved when the Cpu and Gpu testing subfolders were created. Moved small 911 test into Cpu subfolder since it was generated with Cpu code. This required corresponding update to github actions. Then updated RunTests so that it would build the appropriate list of tests based on the current processing unit.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


NicolasJPosey:
@stiber this should be a quick review, but it is ready when you get a chance. I'll then set up a PR to merge this into Shared Development.

stiber (approved):

PR #807, opened 2025-03-03
[issue-806] Implement integrateVertexInputs in All911Vertices

### LINKED ISSUE #806: Implement All911Vertices::integrateVertexInputs
During the refactor of the GPU model, we created a new method called integrateVertexInputs. This was implemented in the neuro model. The 911 model was left alone and a placeholder implementation was created so that the code could be compiled and run while we thought out what logic should move into this method.

NicolasJPosey:
Closes #806

#### Description
Since advanceEdges doesn't update the edge state until after we know that we can transfer the call, we move all of the advanceEdges logic into integrateVertexInputs.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


NicolasJPosey:
@stiber based on the edge state not being updated until after we know that we need to transfer the call to the vertex, I think it's reasonable for all of the current advanceEdges implementation to be moved into integrateVertexInput. Lets discuss this PR at our 1:1 later this week.

stiber (approved):

PR #814, opened 2025-03-07
Merge PoseyDevelopment into SharedDevelopment

### LINKED ISSUE #786: 911 regression config not being run

NicolasJPosey:
Closes #786 #806

- [ ] Added Documentation 
- [ ] Added Unit Tests 


NicolasJPosey:
@stiber changes are mostly the same as the two PRs that we reviewed. Had a merge conflict with a change Diva made to fix a seg fault that I resolved, but it doesn't look like it's in the list of changes. One weird thing I noticed was that Raiju is causing some of the GPU regression tests to fail while Otachi is not. I thought I remember us running into that last year when I was setting up the GPU test outputs. I think that is also the last time I used Raiju for GPU stuff. Not sure if that is something worth us looking into or not.

stiber (approved):
You'll probably will need to merge `SharedDevelopment` into this first.

PR #816, opened 2025-03-12
[ISSUE 386] Check for local log4cplus file in Home Directory.

### LINKED ISSUE #386: Allow users to have a local log4cplus_configure.ini file for changing logging levels
The current log4cplus_configure.ini file under `build/RuntimeFiles` is under version control by `git`. This means that when changing its settings for debugging purposes, you run the risk of committing and pushing those changes upstream if you don't take care to exclude the file from your commits.

An option would be to allow users to have a local `log4cplus_configure.ini` file in a specific directory. The Simulator would check for the existence of the local file, and only use the default one under `build/RuntimeFiles` if it doesn't exist. This will allow users to change their logging level without modifying a file that is under version control.

PadPatil:
Closes #386 

#### Description
This will check for the user's home directory for an existing log4cplus file and it will configure it. Otherwise, the existing RuntimeFiles log4cplus file will be configured.

- [ ] Added Documentation 
- [ ] Added Unit Tests


stiber (approved):
Looks good!

PR #823, opened 2025-03-24
Recorder Updates for Neuro and NG911

### LINKED ISSUE #574: Enhance Recordable Class to Support Variables for NG911 Simulation
So far, we have considered only recording events which is the variable of Neuron network simulation so the only thing that needs to be recorded is the time. However, there is a need to extend the functionality of the Recordable class to accommodate the recording of values for NG911 simulation, enabling the capture of time-series data comprising both time and corresponding values(may be pair?).

### LINKED ISSUE #719: Move the NG911 simulation to the new recorders
The `Xml911Recorder` class currently inherits from `XmlRecorder`, which primarily functions as a buffer for data and outputs text. We need to modify the `Xml911Recorder` class's data members to inherit from `Recordable` instead, enabling compatibility with the updated recording system.

### LINKED ISSUE #722: Double check the default setup for what variables are recorded in our simulation code
For growth, STDP, and 911 simulations. Also, make sure this includes both variables that need to be recorded each epoch and a good set of variables that don't change during the simulation (parameters and initial values).

zshaikh5:
Closes #722 
Closes #719 
Closes #574 

#### Description
This expands the set of variables recorded by including additional ones, providing a wider range of the data captured. Additionally, it removes the dependency on XML911Recorder by using the recorders instead.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


stiber (changes requested):
Before we can merge this, we need to manually check the various new `GoodOutput` files against the contents of the same files in `SharedDevelopment`. @stiber will lead that.

NicolasJPosey (approved):

PR #830, opened 2025-04-10
STDP code clean-up and fixes

### LINKED ISSUE #829: STDP Long Term Depression code is not executing
Running the existing implementation, the resulting edge weights all go to the positive weight max (Wex). Expected behavior is to see an asymmetric weight matrix, where one edge between two neurons approaches the max weight and the edge in the reverse direction approaches zero. 

All edges are approaching the max because the Long Term Depression (LTD) code is not being executed. Investigate the break conditions to make sure they are set up correctly

varndorfer:
Closes #829

#### Description
1. General code clean up including adding comments, removing variables that are declared but never used, and renaming confusing variable names.
2. DynamicSTDP code wasn't working. Found that the DeltaT parameter was mistakenly sent in the setup function and fixed it to send the correct TotalVertices value instead.
3. The STDP function was increasing all weights to the max weight value (Wex). Investigated and found that LTD code wasn't being hit due to an incorrect break condition when checking against STDPgap.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


stiber (approved):
Great job!!

Issue #825, opened 2025-04-05
Refactor copyToGPU into separate allocateOnGPU and copyToGPU methods


BenYang2002:
The current copyToGPU method is responsible for both allocating memory on the GPU and copying data into that memory. To improve modularity, readability, and reusability, we propose splitting this function into two distinct methods:

allocateOnGPU() – responsible for memory allocation only

copyToGPU() – responsible for copying data to the already allocated GPU memory

This change will make GPU memory operations more flexible and easier to manage independently, especially for cases where allocation and copying need to be handled separately (e.g., zero-copy memory sharing, conditional data updates, or deferred initialization).

Tasks:

 Extract allocation logic into allocateOnGPU()

 Extract copy logic into copyToGPU()

 Update existing usages of copyToGPU() to use the two new methods in sequence

 Add comments and documentation for the two new methods

 Add tests

Issue #821, opened 2025-03-20
Add in registerHistoryVariables to OperationManager


stiber:
Will replace explicit class-by-class calls in `Core.cpp`:

https://github.com/UWB-Biocomputing/Graphitti/blob/62ad162a8472397f2a5436f0a9a52da23c4f2f7b/Simulator/Core/Core.cpp#L193-L194

Issue #131, opened 2020-10-23
Implement remaining operations


stiber:
Refer to Operations.h. Implement the remaining operations in this order:

- [x] copyFromGPU
- [x] deallocateGPUMemory
- [x] copyToGPU

stiber:
Includes #130 and #122.

stiber:
Here's a note from [PR 693](https://github.com/UWB-Biocomputing/Graphitti/pull/693#discussion_r1747538637):

Before this change, `registerSpikeHistoryVariable` (now `registerHistoryVariables`) was a member function of `AllSpikingNeurons` and was called within `AllSpikingNeurons::setupVertices()`.

The main reason for moving this function to the AllVertices base class is to accommodate deserialization. Previously, our flow was `setup -> register -> deserialize -> simulate`, and recording variables during the setup stage meant missing out on the deserialized vertices.

With this change, the new flow is `setup -> deserialize -> register -> simulate`, ensuring that all deserialized vertices, are properly recorded.

stiber:
Also, need to incorporate things done collectively by `Simulator::setup()`, `Model::sertupSim()`Layout::setup()`, `Vertices:: setupVertices()`, etc., and their subclasses.

BenYang2002:
I’d like to clarify the role of the current OperationManager and why it may not be suitable for the remaining three operations (copyFromGPU, deallocateGPUMemory, copyToGPU), as listed in this issue.

The existing OperationManager is often referred to in documentation as implementing the Chain of Responsibility (CoR) pattern, but after reviewing its design, it's more accurate to describe it as a central dispatcher. It allows various components to register callbacks under a shared operation type, and when an operation is triggered (via executeOperation), all matching callbacks are executed. This is closer to a pub-sub or broadcast mechanism, not a true CoR.

In contrast, the remaining three GPU-related operations exhibit a very different pattern. The implementation of GPUModel::allocDeviceStruct() and related methods shows a clear hard-coded chain of steps where memory is allocated and transferred across subsystems like neurons and synapses in a specific sequence. This is where an actual Chain of Responsibility would make sense: each component performs its operation and passes control to the next in a flexible and modular fashion.

Therefore, I suggest we introduce a new class (e.g., ChainManager) specifically for modeling this kind of sequential operation flow. This will allow us to:

Encapsulate each GPU-related step as a handler

Make the flow easier to test and modify

Avoid polluting OperationManager, which serves a different purpose

[...]

PR #841, opened 2025-04-24
[ISSUE-131] Moving GPU memory management to OperationManager

BenYang2002:
Closes
 [#131](https://github.com/UWB-Biocomputing/Graphitti/issues/131)
#825 
#821 

#### Description
Summary
This PR completes the integration of the remaining GPU operations — copyFromGPU and deallocateGPUMemory — into the OperationManager framework, following the previously refactored copyToGPU. All related methods have been standardized to take no parameters and are now registered as std::function<void()> callbacks. This centralizes GPU operation dispatch, eliminates hardcoded call chains, and improves modularity and maintainability.

Changes
Refactored copyFromGPU and deallocateGPUMemory to be parameterless

Moved required context into member variables

Registered operations during class construction

Added #ifdef USE_GPU guards where applicable

Replaced explicit calls in GPUModel with OperationManager::execute(...)

Maintained consistency with the existing copyToGPU refactor

- [x] Added Documentation 
- [] Added Unit Tests 


stiber:
This PR isn't named properly, and it doesn't indicate which issue(s) it closes.

stiber (changes requested):
I need more information about this PR. Which issues is it connected to? It seems to include multiple ones.

Simulator/Core/GPUModel.cpp
```diff
@@ -25,30 +26,31 @@ GPUModel::GPUModel() :
    Model::Model(), synapseIndexMapDevice_(nullptr), randNoise_d(nullptr),
    allVerticesDevice_(nullptr), allEdgesDevice_(nullptr)
 {
+#if defined(USE_GPU)
```
stiber, line 29: A couple questions:

1. Is the conditional compilation needed for GPUModel?
2. Where is CopyFromGPU?

Simulator/Core/GPUModel.cpp
```diff
@@ -25,30 +26,31 @@ GPUModel::GPUModel() :
...
+   // Register allocNeuronDeviceStruct function as a allocateGPU operation in the OperationManager
+   function<void()> allocateGPU = bind(&GPUModel::allocDeviceStruct, this);
+   OperationManager::getInstance().registerOperation(Operations::allocateGPU, allocateGPU);
+
+   // Register copySynapseIndexMapHostToDevice function as a copyCPUtoGPU operation in the OperationManager
+   function<void()> copyCPUtoGPU = bind(&GPUModel::copySynapseIndexMapHostToDevice, this);
```
stiber, line 35: The name of this function doesn't seem to match what is defined in the class, which is `virtual void copyCPUtoGPU() override;`

Simulator/Core/GPUModel.cpp
```diff
@@ -62,14 +64,10 @@ void GPUModel::deleteDeviceStruct(void **allVerticesDevice, void **allEdgesDevic
    AllVertices &neurons = layout_->getVertices();
    AllEdges &synapses = connections_->getEdges();
 
-   // Copy device synapse and neuron structs to host memory
-   neurons.copyFromDevice(*allVerticesDevice);
    // Deallocate device memory
-   neurons.deleteNeuronDeviceStruct(*allVerticesDevice);
-   // Copy device synapse and neuron structs to host memory
-   synapses.copyEdgeDeviceToHost(*allEdgesDevice);
+   neurons.deleteNeuronDeviceStruct();
```
stiber, line 68: Why doesn't the vertex class register this method with the operation manager?

Simulator/Core/GPUModel.cpp
```diff
@@ -62,14 +64,10 @@ void GPUModel::deleteDeviceStruct(void **allVerticesDevice, void **allEdgesDevic
...
    // Deallocate device memory
-   synapses.deleteEdgeDeviceStruct(*allEdgesDevice);
+   synapses.deleteEdgeDeviceStruct();
```
stiber, line 70: Why doesn't the edge class register this method with the operation manager?

Simulator/Core/GPUModel.cpp
```diff
@@ -120,9 +116,10 @@ void GPUModel::setupSim()
 /// Performs any finalization tasks on network following a simulation.
 void GPUModel::finish()
 {
+   // copy device synapse and neuron structs to host memory
+   copyGPUtoCPU();
```
stiber, line 120: Shouldn't this be handled by the OperationManager?

Simulator/Core/GPUModel.cpp
```diff
@@ -198,7 +195,7 @@ void GPUModel::updateConnections()
    AllVertices &neurons = layout_->getVertices();
    AllEdges &synapses = connections_->getEdges();
 
-   dynamic_cast<AllSpikingNeurons &>(neurons).copyFromDevice(allVerticesDevice_);
+   dynamic_cast<AllSpikingNeurons &>(neurons).copyFromDevice();
```
stiber, line 198: We should go over this code, both from the point of view of the dynamic cast and whether the OperationManager should be used here.

Simulator/Core/GPUModel.cpp
```diff
@@ -210,8 +207,7 @@ void GPUModel::updateConnections()
       // create synapse index map
       connections_->createEdgeIndexMap();
       // copy index map to the device memory
-      copySynapseIndexMapHostToDevice(connections_->getEdgeIndexMap(),
-                                      Simulator::getInstance().getTotalVertices());
+      copySynapseIndexMapHostToDevice();
```
stiber, line 210: We should go over this code, both from the point of view of why this is a GPUModel method and  whether the OperationManager should be used here.

Simulator/Core/GPUModel.cpp
```diff
@@ -263,8 +259,10 @@ void GPUModel::deleteSynapseImap()
 
 /// Copy EdgeIndexMap in host memory to EdgeIndexMap in device memory.
 /// @param  synapseIndexMapHost		Reference to the EdgeIndexMap in host memory.
-void GPUModel::copySynapseIndexMapHostToDevice(EdgeIndexMap &synapseIndexMapHost, int numVertices)
+void GPUModel::copySynapseIndexMapHostToDevice()
```
stiber, line 262: Why is this is a GPUModel method?

Simulator/Core/GPUModel.cpp
```diff
@@ -362,19 +360,34 @@ __global__ void
 /// Copy GPU Synapse data to CPU.
 void GPUModel::copyGPUtoCPU()
 {
-   // copy device synapse structs to host memory
-   connections_->getEdges().copyEdgeDeviceToHost(allEdgesDevice_);
+   // copy device neuron and synapse structs to host memory
+   OperationManager::getInstance().executeOperation(Operations::copyFromGPU);
```
stiber, line 364: This seems circular. Isn't it called by the OperationManager?

Simulator/Core/GPUModel.cpp
```diff
@@ -362,19 +360,34 @@ __global__ void
...
 /// Copy CPU Synapse data to GPU.
 void GPUModel::copyCPUtoGPU()
 {
-   // copy host synapse structs to device memory
-   connections_->getEdges().copyEdgeHostToDevice(allEdgesDevice_);
+   // copy host neurons and synapse structs to device memory
+   AllVertices &neurons = layout_->getVertices();
+   AllEdges &synapses = connections_->getEdges();
+   neurons.copyToDevice();
+   synapses.copyEdgeHostToDevice();
```
stiber, line 374: Shouldn't these be called by the OperationManager?

stiber (changes requested):
I just looked at a couple files and I have many questions. I think that this PR is just too big and complex. This should have been done in smaller chunks, and/or it should have been made earlier so that interim code reviews could have been done. So, I will stop with 2 out of 26 files reviewed and around 10 questions. Once we get through that, we can move on to the next set.

BenYang2002:
> This PR isn't named properly, and it doesn't indicate which issue(s) it closes.

I just included the issues it closes, also for the issue131(implementing the remaining operations) I included the subissues for it

Simulator/Core/GPUModel.cpp
BenYang2002: 1.No, it’s not needed. I initially added it as a safeguard, but the model includes it during construction. I’ve removed the flag.
2.GPUModel doesn’t register CopyFromGPU — neurons and synapses handle copying results back. The model itself has nothing to copy.

Simulator/Core/GPUModel.cpp
BenYang2002: The original copyCPUtoGPU() in the master branch actually copies device synapse structs back to host, which is the synapses' responsibility — not the model’s. Also, that function is never invoked anywhere.
The actual GPUModel-level logic for copying from GPU to CPU is in copySynapseIndexMapHostToDevice(), which is why I’m registering that function instead.

Simulator/Core/GPUModel.cpp
BenYang2002: The original deleteDeviceStruct() function included code that is no longer needed after the refactor. This function was previously unused, but has now been repurposed and is invoked via finish().

New implementation:
/// Copies device memories to host memories and deallocates them.
void GPUModel::deleteDeviceStruct()
{
    OperationManager::getInstance().executeOperation(Operations::deallocateGPUMemory);
}
This avoids calling OperationManager::executeOperation(...) directly inside finish(), making the deallocation step modular and consistent with the updated GPU operation framework.

Simulator/Core/GPUModel.cpp
BenYang2002: Both synapse and neuron deallocation methods are now registered with the OperationManager.
The old deleteDeviceStruct() function became unused after the refactor and effectively turned into a zombie.
Instead of keeping it, I replaced its logic with a single call:
OperationManager::getInstance().executeOperation(Operations::deallocateGPUMemory);
This call is now made directly from the finish() function.

Simulator/Core/GPUModel.cpp
BenYang2002: The copyGPUtoCPU() function was effectively just a single line : OperationManager::getInstance().executeOperation(Operations::copyFromGPU);
Given that, the finish() function are actually doing:
OperationManager::getInstance().executeOperation(Operations::copyFromGPU);
OperationManager::getInstance().executeOperation(Operations::deallocateGPUMemory);

Simulator/Core/GPUModel.cpp
BenYang2002: From the code change perspective: the only change is that now neuron.copyFromDevice takes in no parameter.
No comment on the dynamic cast
For use of OperationManager, we cannot directly use copyGPUToCPU, since what it does is copying everything: neuron,synapses and data in GPUModel back to CPU. So if we want to use OperationManager just for copy neuron data, we will have to make and register sub operation just for neuron

Simulator/Core/GPUModel.cpp
BenYang2002: Similar with the previous conversation, the changing part is that now copySynapseIndexMapHostToDevice takes no parameter

Simulator/Core/GPUModel.cpp
BenYang2002: I will look into the logic and determine which class should have it instead. All I am doing right now is just refactoring such that it takes no parameter for registering purpose

Simulator/Core/GPUModel.cpp
BenYang2002: The copyFromGPU operation is used to trigger the actual copying of data from GPU to CPU, and different components (like neurons and synapses) register their own logic to this operation via the OperationManager.

Calling GPUModel::copyGPUtoCPU() simply tells the OperationManager to execute all registered copyFromGPU callbacks.
Those callbacks handle their local data transfer, and none of them call copyGPUtoCPU(), so there’s no circular behavior.

Simulator/Core/GPUModel.cpp
BenYang2002: this is actually another zombie function that's never used. now the copyCPUToGPU is replaced with:
void GPUModel::copyCPUtoGPU()
{
   // Allocate and copy neuron/synapse data structures to GPU memory
   OperationManager::getInstance().executeOperation(Operations::allocateGPU);
   // Copy host neuron and synapse arrays into GPU device
   // copy inverse map to the device memory
   OperationManager::getInstance().executeOperation(Operations::copyToGPU);
}

Simulator/Core/GPUModel.cpp
BenYang2002: The EdgeIndexMap is an indexing data structure used to efficiently access synapse connections for each neuron during GPU computation. It maintains flat arrays of outgoing and incoming edge indices, along with per-neuron offset and count arrays (*_Begin_, *_Count_) to enable fast slicing. This allows kernels like calcSummationPointDevice() to quickly sum inputs from incoming synapses without scanning the entire synapse array. Since the map only stores static structural information (not simulation results) and is rebuilt on the CPU when connections change, it does not need to be copied back from GPU to CPU.

Simulator/Core/GPUModel.cpp
BenYang2002: issue resolved, now finish function looks like this:
void GPUModel::finish()
{
   OperationManager::getInstance().executeOperation(Operations::copyFromGPU);
   OperationManager::getInstance().executeOperation(Operations::deallocateGPUMemory);

#ifdef PERFORMANCE_METRICS
   cudaEventDestroy(start);
   cudaEventDestroy(stop);
#endif   // PERFORMANCE_METRICS
}

Simulator/Core/GPUModel.cpp
```diff
@@ -27,51 +29,46 @@ GPUModel::GPUModel() :
...
    allEdgesDevice_(nullptr)
 {
+   // Register allocNeuronDeviceStruct function as a allocateGPU operation in the OperationManager
+   function<void()> allocateGPU = bind(&GPUModel::allocDeviceStruct, this);
+   OperationManager::getInstance().registerOperation(Operations::allocateGPU, allocateGPU);
+
+   // Register copyCPUtoGPU function as a copyCPUtoGPU operation in the OperationManager
+   function<void()> copyCPUtoGPU = bind(&GPUModel::copyCPUtoGPU, this);
+   OperationManager::getInstance().registerOperation(Operations::copyToGPU, copyCPUtoGPU);
+
```
stiber, line 39: Add a comment indicating why no copy from GPU.

Simulator/Core/Serializer.cpp
```diff
@@ -67,8 +67,7 @@ bool Serializer::deserialize()
 
 #if defined(USE_GPU)
    GPUModel &gpuModel = static_cast<GPUModel &>(simulator.getModel());
-   gpuModel.copyEdgeIndexMapHostToDevice(simulator.getModel().getConnections().getEdgeIndexMap(),
-                                         simulator.getTotalVertices());
+   gpuModel.copyCPUtoGPU();
```
stiber, line 70: Need to ask @d-kamath : why is only the `GPUModel` data being copied to the GPU? Why aren't we copying everything (which can be accomplished via the `OperationManager` now)?

Testing/test_output.log
stiber, line 1: Remove file. Check that .gitignore prevents this from being added to repo in future.

stiber (changes requested):
Some comments.

Simulator/Core/GPUModel.h
```diff
@@ -92,20 +93,24 @@ class GPUModel : public Model {
    /// over the past epoch. Should be called once every epoch.
    virtual void updateConnections() override;
 
-   /// Copy GPU Synapse data to CPU.
+   /// Copy GPU Neuron and Synapse data to CPU.
    virtual void copyGPUtoCPU() override;
 
-   /// Copy CPU Synapse data to GPU.
+   /// Copy CPU Neuron and Synapse data to GPU.
```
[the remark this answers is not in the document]
BenYang2002: added the comment , and gonna make an issue for this

Simulator/Core/GPUModel.h
```diff
@@ -92,20 +93,24 @@ class GPUModel : public Model {
...
 
    /// Print out SynapseProps on the GPU.
    void printGPUSynapsesPropsModel() const;
 
+   /// Getter for synapse structures in device memory
+   AllSpikingSynapsesDeviceProperties *&getAllEdgesDevice();
+
+   /// Getter for neuron structure in device memory
+   AllSpikingNeuronsDeviceProperties *&getAllVerticesDevice();
+
```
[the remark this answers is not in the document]
BenYang2002: added the comments

Simulator/Core/Operations.h
```diff
@@ -22,6 +22,8 @@ class Operations {
       deallocateGPUMemory,   // Make sure deallocate memory isn't called until all GPU memory is copied back.
       restoreToDefault,   // Not sure what this refers to.
       copyToGPU,
-      copyFromGPU
+      copyFromGPU,
+      allocateGPU,
+      registerHistoryVariables
```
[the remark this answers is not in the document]
BenYang2002: I think some other developer added the registerHistoryVariables

Simulator/Core/Serializer.cpp
```diff
@@ -67,8 +67,7 @@ bool Serializer::deserialize()
 
 #if defined(USE_GPU)
    GPUModel &gpuModel = static_cast<GPUModel &>(simulator.getModel());
-   gpuModel.copySynapseIndexMapHostToDevice(simulator.getModel().getConnections().getEdgeIndexMap(),
-                                            simulator.getTotalVertices());
+   gpuModel.copySynapseIndexMapHostToDevice();
```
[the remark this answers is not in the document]
BenYang2002: I contacted Dyvia and will post her answer later she replied

Simulator/Core/GPUModel.cpp
BenYang2002: added a comment:
   // Note: We do not register a corresponding copyFromGPU operation here because
   // we are only copying the synapseIndexMap to the GPU. This map is a read-only lookup table
   // that gets recreated from scratch on each update. As a result, we only need to allocate,
   // copy to GPU, and deallocate — there is no meaningful data to copy back from the GPU.

Simulator/Core/Serializer.cpp
BenYang2002: I contacted Dyvia and will post her answer later she replied

Testing/test_output.log
BenYang2002: removed the file

PR #850, opened 2025-05-12
Device Vector added and Integrated with Neuro Vertices

Avikantw:
#### Description
1. DeviceVector.h is added in the Utils and integrated with neural vertex classes. It is conditionally compiled on the basis of __CUDACC__ macros defined by nvcc. 

2. Structures are basically empty now except it still holds only event buffer for the device code.

3. Each kernel takes explicit parameter.

5. Units tests are also added which tests both CPU and GPU functionality for device vector.

- [x] Added Documentation 
- [x] Added Unit Tests 

### DIFF
--- a/Simulator/Utils/DeviceVector.h
+++ b/Simulator/Utils/DeviceVector.h
@@ -1,0 +1,455 @@
+/**
+ * @file DeviceVector.h
+ *
+ * @ingroup Simulator/Utils
+ *
+ * @brief A Template class that provides a std::vector-like interface for both host and device (GPU) memory
+ * @tparam T Type of elements (must be BGFLOAT, int, or bool)
+ *
+ * This class wraps a std::vector and provides additional operations for device memory
+ * allocation, deallocation, and host-device data transfer. Currently, it only supports
+ * primitive types, and since the codebase only uses BGFLOAT, int, and bool, these are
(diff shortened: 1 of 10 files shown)


stiber (changes requested):
Mostly commenting. Some questions.

Simulator/Utils/DeviceVector.h
```diff
@@ -0,0 +1,439 @@
...
+
+/// Define supported types for DeviceVector using variant
+using DeviceVectorTypes = std::variant<BGFLOAT, int, bool>;
+
+/// Type trait to check if a type is one of the alternatives in a variant
+template <typename T, typename Variant> struct is_device_vector_type : std::false_type {};
+
+template <typename T, typename... Types>
+struct is_device_vector_type<T, std::variant<Types...>> :
+   std::bool_constant<(std::is_same_v<T, Types> || ...)> {};
```
stiber, line 59: I hate to do this, but could you include more explanation of this in the comment? For example, the obvious first question is why doesn't this trigger a "multiply defined" compiler error for `struct is_device_vector_type`?

stiber (changes requested):
One small comment.

stiber (approved):

PR #859, opened 2025-06-10
[ISSUE-815] Generate Random Numbers Asynchronously on the GPU

### LINKED ISSUE #815: Generate random numbers asynchronously on the GPU
ggraphitti's current random number generation system consists of a series of C function calls, with RandomNormGPU as the kernel that is repeatedly invoked after the initial setup. This approach generates only a small number of random numbers per call, resulting in low GPU occupancy and frequent kernel launches. My plan is to introduce a dedicated class that generates random numbers asynchronously and in larger batches, improving GPU utilization while also enhancing encapsulation and modularity.

AndrewBMadison:
Closes #815

#### Description
Replaced the custom Mersenne Twister GPU kernel with an AsyncPhilox_d class that asynchronously fills GPU buffers with random noise using cuRAND's Philox generator. The class supports double-buffering and is designed for concurrent execution.

GPUModel initializes Philox states and fills two initial buffers via loadPhilox() on a member AsyncPhilox_d instance. During each advance() call, requestSegment() retrieves a float* slice from the currently active buffer, sized appropriately for each vertex and ready to be used in advanceVertices(). 

Once a buffer is consumed, fillBuffer() is triggered on the other buffer while the current one continues to serve slices. This ensures continuous data availability through double-buffering.

AsyncPhilox_d uses its own internal CUDA stream to launch fill kernels asynchronously. To enable true concurrency, all other compute kernels needed to also use non-default streams. This is necessary because stream 0 (the default stream) implicitly synchronizes with all other streams, preventing concurrent execution and causing the scheduler to serialize kernel launches even when they could run in parallel.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


Simulator/Connections/Connections.h
```diff
@@ -87,7 +92,8 @@ class Connections {
    ///  @param  layout               Layout information of the graph network.
```
stiber, line 92: Please update the comment for this method to reflect current set of parameters.

Simulator/Connections/Neuro/ConnGrowth.h
```diff
@@ -127,8 +127,9 @@ class ConnGrowth : public Connections {
    ///  @param  layout               The Layout object.
```
stiber, line 127: Please update the comment for this method to reflect current set of parameters.

Simulator/Edges/AllEdges.cpp
```diff
@@ -217,9 +217,16 @@ void AllEdges::createEdgeIndexMap(EdgeIndexMap &edgeIndexMap)
    }
 }
 
+#if defined(USE_GPU)
+// Set GPU stream for edge kernels
```
stiber, line 221: Document a bit better in comment.

Simulator/Utils/RNG/MersenneTwister_d.h
stiber, line 1: Is this still used?

Simulator/Vertices/AllVertices.cpp
```diff
@@ -49,3 +49,12 @@ void AllVertices::loadEpochInputs(uint64_t currentStep, uint64_t endStep)
    // This is an empty implementation so that Neural Network simulation works
    // normally
 }
+
+
+#ifdef USE_GPU
+// Set Cuda Stream for Vertices kernels
+void AllVertices::SetStream(cudaStream_t pStream)
+{
+   stream = pStream;
```
stiber, line 58: `stream` is a data member of the class; should be named `stream_`. Is this supposed to be distinct from the `AllEdges` stream?

Simulator/Vertices/AllVertices.h
```diff
@@ -92,7 +95,13 @@ class AllVertices {
    log4cplus::Logger vertexLogger_;   // Logs to Output/Debug/neurons.txt
 
 #if defined(USE_GPU)
+   ///  Cuda Stream for Vertices Kernels
+   cudaStream_t stream;
```
stiber, line 99: `stream` is a data member of the class; should be named `stream_`. Is this supposed to be distinct from the `AllEdges` stream?

Simulator/Vertices/Neuro/AllIZHNeurons_d.cpp
```diff
@@ -171,7 +171,7 @@ void AllIZHNeurons::advanceVertices(AllEdges &synapses, void *allVerticesDevice,
    int blocksPerGrid = (vertex_count + threadsPerBlock - 1) / threadsPerBlock;
 
    // Advance neurons ------------->
-   advanceIZHNeuronsDevice<<<blocksPerGrid, threadsPerBlock>>>(
+   advanceIZHNeuronsDevice<<<blocksPerGrid, threadsPerBlock, 0, stream>>>(
```
stiber, line 174: `stream` is a data member of the class; should be named `stream_`. Is this supposed to be distinct from the `AllEdges` stream?

stiber (changes requested):
Minor cleanup.

Simulator/Edges/Neuro/AllSTDPSynapses_d.cpp
```diff
@@ -268,7 +268,7 @@ void AllSTDPSynapses::advanceEdges(void *allEdgesDevice, void *allVerticesDevice
    const int threadsPerBlock = 256;
    int blocksPerGrid = (totalEdgeCount_ + threadsPerBlock - 1) / threadsPerBlock;
    // Advance synapses ------------->
-   advanceSTDPSynapsesDevice<<<blocksPerGrid, threadsPerBlock>>>(
+   advanceSTDPSynapsesDevice<<<blocksPerGrid, threadsPerBlock, 0, stream>>>(
```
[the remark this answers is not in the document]
AndrewBMadison: The new name for the stream is simulationStream and it is the same stream used by all synchronous kernels.

Simulator/Vertices/AllVertices.cpp
AndrewBMadison: The new name for the stream is simulationStream and it is the same stream used by all synchronous kernels.

Simulator/Vertices/AllVertices.h
AndrewBMadison: The new name for the stream is simulationStream and it is the same stream used by all synchronous kernels.

Simulator/Vertices/Neuro/AllSpikingNeurons_d.cpp
```diff
@@ -194,7 +194,7 @@ void AllSpikingNeurons::integrateVertexInputs(void *allVerticesDevice,
       = (Simulator::getInstance().getTotalVertices() + threadsPerBlock - 1) / threadsPerBlock;
    int vertex_count = Simulator::getInstance().getTotalVertices();
 
-   calcSummationPointDevice<<<blocksPerGrid, threadsPerBlock>>>(
+   calcSummationPointDevice<<<blocksPerGrid, threadsPerBlock, 0, stream>>>(
```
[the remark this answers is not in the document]
AndrewBMadison: The new name for the stream is simulationStream and it is the same stream used by all synchronous kernels.

Simulator/Utils/RNG/MersenneTwister_d.h
AndrewBMadison: It is no longer used, but I had left is in in case someone wanted to uncomment the code that uses the old RNG and test it. If you would like I could remove these files.

stiber (dismissed):
Looks great; I will merge this.

stiber:
I take it back; this needs to have SharedDevelopment merged into it. There may be a conflict with the changes to device memory allocation/deallocation being moved to `OperationManager`. We will need to let all the tests run, plus I will add a comment to one location in the code.

Simulator/Core/GPUModel.cpp
```diff
@@ -53,28 +53,31 @@ void GPUModel::allocDeviceStruct()
...
-   HANDLE_ERROR(cudaFree(edgeIndexMapDevice_));
-   HANDLE_ERROR(cudaFree(randNoise_d));
+   vertices.deleteVerticesDeviceStruct(*allVerticesDevice);
+   // Copy device edge and vertex structs to host memory
+   edges.copyEdgeDeviceToHost(*allEdgesDevice);
+   // Deallocate device memory
+   edges.deleteEdgeDeviceStruct(*allEdgesDevice);
+   // HANDLE_ERROR(cudaFree(randNoise_d));
+   // closeFileMT();
+   HANDLE_ERROR(cudaStreamDestroy(simulationStream_));
```
stiber, line 80: This method needs review. Questions include:
1. Where are `vertices` and `edges` declared?
2. Given the new `OperationManager` control that has all classes managing their own device memory, are any of these "non-local" deletions needed?

Simulator/Core/GPUModel.cpp
```diff
@@ -124,8 +133,11 @@ void GPUModel::finish()
    // copy device synapse and neuron structs to host memory
    OperationManager::getInstance().executeOperation(Operations::copyFromGPU);
    // deallocates memories on CUDA device
+   // TODO: Resolve whether the deletion operations below that don't seem bundled
+   // with the OperationManager are necessary, should be moved, or if they can be removed.
+   AsyncGenerator_.deleteDeviceStruct();
    OperationManager::getInstance().executeOperation(Operations::deallocateGPUMemory);
-
+   deleteEdgeIndexMap();
```
stiber, line 140: As the TODO says, the two lines that delete `AsyncGenerator_` and the edge index map either need to be moved or aren't necessary.

Simulator/Core/GPUModel.cpp
```diff
@@ -238,7 +251,8 @@ void GPUModel::updateConnections()
    // Update Connections data
    if (connections_->updateConnections(vertices)) {
       connections_->updateEdgesWeights(Simulator::getInstance().getTotalVertices(), vertices, edges,
-                                       allVerticesDevice_, allEdgesDevice_, getLayout());
+                                       allVerticesDevice_, allEdgesDevice_, getLayout(),
+                                       simulationStream_);
       // create edge index map
       connections_->createEdgeIndexMap();
       // copy index map to the device memory
```
stiber, line 258: Should the line below use `OperationManager`, or do we want finer granularity?

stiber (changes requested):
Besides the comments below, need to examine `GPUModel::allocEdgeIndexMap()` and `GPUModel::copyCPUtoGPU()` to see if they need rewrites because of `DeviceVector`

Simulator/Core/GPUModel.cpp
AndrewBMadison: I have moved the deletion of AsyncGenerator_ to deleteDeviceStruct(), and other line is now deleted from the remerge of SharedDevelopment.

Simulator/Core/GPUModel.cpp
AndrewBMadison: That is a good question and it looks like OperationManager could be used to execute this operation, but I don't know if I'm the right person to answer that question.

AndrewBMadison:
I reverted and remerged SharedDevelopment into AndrewDevelopment, manually reviewing changes and deleting old unnecessary code from the SharedDevelopment commit before Ben merged his code in. This resolved a lot of the changes you requested, but I am unsure if the OperationManager should execute the copyCPUtoGPU so maybe you could ask Ben. I also moved the AsyncGenerator deletion as you requested.

PR #871, opened 2025-08-07
Release v 1.1

stiber:
Release for summer 2025, ready for final checks and release creation.


NicolasJPosey:
@stiber I was seeing some compile issues on a branch I just made off of SharedDevelopment when compiling for the GPU so I checked this branch out to confirm if it was something I did or not. I got the same errors. They are as follows:
1. When compiling, an error is detected in ConnGrowth_d.cpp. We are calling `layout.vertexTypeMap_.data()`. The map was recently changed from a vector to a RecordableVector but RecordableVector doesn't implement `std::vector.data()`. I locally tried implementing `T *data()` in RecordableVector and this resolved the issue. We are using it in a HostToDevice copy so maybe `const T *data()` is the better one to implement.
2. After resolving 1, I get a error in serialFirstHalfTest (I think it's during the linkage step). It says that there is an undefined reference to Connections911::registerHistoryVariables(). I don't get this error often enough to know what exactly is needed to resolve it.

NicolasJPosey:
@stiber I figured out 2. The registerHistoryVariables implementation was inside of a #if !(USE_GPU) so the linker couldn't find it. I will create a branch off of SharedDevelopment with these changes and we can pull them into the release branch when we are happy with the proposed fixes.

NicolasJPosey (approved):

Issue #870, opened 2025-08-07
Clean up EventBuffer implementation


stiber:
There are a number of methods that override `RecordableVector` methods but don't need to; the inherited ones would work fine (and sometimes the override is identical).

NicolasJPosey:
Doing this clean with #873

NicolasJPosey:
Close as implemented in PR 876

PR #876, opened 2025-08-10
[issue-873] Templated Event Buffer

### LINKED ISSUE #873: Upgrade EventBuffer to be templated
The current implementation of the EventBuffer implements a RecordableVector of type uint_64t. It doesn't look like there is any reason the EventBuffer couldn't be turned into an EventBuffer<T> so that it can be used with other types. This would be immediately useful in the NG911 implementation.

NicolasJPosey:
Closes #873 #870 

#### Description
- Upgraded EventBuffer to be a templated class so that it could be used with other types.
- Cleaned up redundant overrides in EventBuffer .
- Added getters, setters, and a couple of methods overrides like size in EventBuffer and data in RecordableVector so that the friend coupling to neuro classes in EventBuffer could be removed.

- [ ] Added Documentation 
- [ ] Added Unit Tests 


Simulator/Vertices/EventBuffer.h
```diff
@@ -41,7 +36,13 @@ class EventBuffer : public RecordableVector<uint64_t> {
    /// an empty and a full buffer.
    ///
    /// @param maxEvents Defaults to zero; otherwise, buffer size is set
-   EventBuffer(int maxEvents = 0);
+   EventBuffer(int maxEvents = 0)
+   {
+      //RecordableVector<T>();
```
stiber, line 41: Please delete this commented out code.

Simulator/Vertices/EventBuffer.h
```diff
@@ -41,7 +36,13 @@ class EventBuffer : public RecordableVector<uint64_t> {
...
+      this->dataSeries_.assign(maxEvents, numeric_limits<T>::max());
```
stiber, line 42: Don't use `this` pointer when unnecessary.

Simulator/Vertices/EventBuffer.h
```diff
@@ -41,7 +36,13 @@ class EventBuffer : public RecordableVector<uint64_t> {
...
+      clear();
+      this->setDataType();   // set up data type for recording purpose
```
stiber, line 44: Ditto on the `this` pointer; will not mention for any future uses, but comment applies to them.

Simulator/Vertices/EventBuffer.h
```diff
@@ -52,31 +53,88 @@ class EventBuffer : public RecordableVector<uint64_t> {
...
    /// in the preceding epoch) is not the same as the number of events in the buffer, because the buffer
    /// retains events from the previous epoch, too.
-   int getNumElementsInEpoch() const;
+   int getNumElementsInEpoch() const
+   {
+      return numElementsInEpoch_;
+   }
+
+   /// Getters are needed for copying from the GPU. Allows us to remove the friend keyword requirement.
+   /// {
```
stiber, line 113: Do you mean "setters" (i.e., mutators)?

stiber (changes requested):
Minor stuff. Let me know if this is essential for this summer's release; otherwise, we will hold it back from SharedDevelopment until the CPU compilation hotfix gets merged into the release.

NicolasJPosey:
@stiber this PR is not essential for this summer's release. I can use the DoubleEventBuffer implementation until this makes it into SharedDevelopment then update the 911 GPU code to use the templated version once SharedDevelopment is merged in.

Simulator/Vertices/EventBuffer.h
NicolasJPosey: It appears to be necessary for this file. When I remove `this->` from the line of code, I get the below error
image
This error appears to happen in every place we try to use the `dataSeries_` member in the file. This member is protected in `RecordableVector<T>`.

I wasn't sure why this is so I asked ChatGPT. It says that it's due to C++'s two-phase name lookup. One solution is to access the base members and methods using `this` pointer. Another is to add a using for each base member and method that we need to access. See below for example.
image

Do you have a preference on using `this` vs adding usings? My preference would be for using `this` since it wouldn't require having to keep adding usings if we need to access more base members or methods in the future.

NicolasJPosey:
@stiber I pulled SharedDevelopment into this branch and reran the unit/regression tests. Everything looks good and is passing. I had resolved the requested changes so I don't know why GitHub isn't just in a pending review state, but this PR is ready for your re-review.

Simulator/Vertices/EventBuffer.h
stiber: Seems reasonable.

stiber (approved):

PR #910, opened 2026-02-08
[ISSUE-7] Moved all model-specific defaults from Global.h to appropriate model classes

### LINKED ISSUE #7: Move model-specific defaults from Global.h to appropriate model class
Right now, there are a bunch of defaults for various parameters of LIF neurons and DS synapses. After checking that it won't break anything, these should be moved.

CodersRepo:
…classes

Closes #7

#### Description
Moved all model-specific defaults from Global.h to appropriate model classes

- [ ] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/Simulator/Edges/Neuro/AllSpikingSynapses.cpp
+++ b/Simulator/Edges/Neuro/AllSpikingSynapses.cpp
@@ -173,7 +173,7 @@ void AllSpikingSynapses::createEdge(BGSIZE iEdg, int srcVertex, int destVertex, 
    sourceVertexIndex_[iEdg] = srcVertex;
    W_[iEdg] = edgSign(type) * 10.0e-9;
    type_[iEdg] = type;
-   tau_[iEdg] = DEFAULT_tau;
+   tau_[iEdg] = AllSpikingSynapses::DEFAULT_tau;
 
    BGFLOAT tau;
    switch (type) {
--- a/Simulator/Edges/Neuro/AllSpikingSynapses.h
+++ b/Simulator/Edges/Neuro/AllSpikingSynapses.h
@@ -38,6 +38,11 @@ using fpPostSynapsesSpikeHit_t = void (*)(const BGSIZE, AllSpikingSynapsesDevice
 
 class AllSpikingSynapses : public AllNeuroEdges {
 public:
+   // Default synapse parameter values
+   static constexpr BGFLOAT DEFAULT_tau = 3e-3;           // The default synaptic time constant
+   static constexpr BGFLOAT DEFAULT_U = 0.4;               // The default synaptic efficiency
+   static constexpr BGFLOAT DEFAULT_delay_weight = 0;     // The default delay weight
+
    AllSpikingSynapses();
 
    AllSpikingSynapses(int numVertices, int maxEdges);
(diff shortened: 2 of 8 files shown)


Simulator/Edges/Neuro/AllSpikingSynapses.h
```diff
@@ -38,6 +38,11 @@ using fpPostSynapsesSpikeHit_t = void (*)(const BGSIZE, AllSpikingSynapsesDevice
 
 class AllSpikingSynapses : public AllNeuroEdges {
 public:
+   // Default synapse parameter values
+   static constexpr BGFLOAT DEFAULT_tau = 3e-3;         // The default synaptic time constant
+   static constexpr BGFLOAT DEFAULT_U = 0.4;            // The default synaptic efficiency
+   static constexpr BGFLOAT DEFAULT_delay_weight = 0;   // The default delay weight
```
stiber, line 44: Why `constexpr` and not just `const`? Just curious.

`DEFAULT_U` should, I think, be a Dynamic Spiking Synapse member.

stiber: Or is the `DEFAULT_U` placement a "feature" of the fact that `AllDynamicSTDPSynapses` doesn't inherit from `AllDSSynapses`?

stiber (changes requested):
Mostly comments, possibly not requiring change.

Simulator/Edges/Neuro/AllSpikingSynapses.h
CodersRepo: > Why `constexpr` and not just `const`? Just curious.

It just made more sense in my head to have them compile at compile-time instead of run-time. constexpr guarantees it, so I used it.

> `DEFAULT_U` should, I think, be a Dynamic Spiking Synapse member.

I will check and get back to you.

CodersRepo: > Or is the `DEFAULT_U` placement a "feature" of the fact that `AllDynamicSTDPSynapses` doesn't inherit from `AllDSSynapses`?

Just quickly double checked. Yes, DEFAULT_U lives in AllSpikingSynapses because it’s shared across multiple spiking synapse models, and AllDynamicSTDPSynapses doesn’t inherit from AllDSSynapses, so this avoids duplication and keeps the default at the appropriate abstraction level. It made sense to keep the code clean this way.

stiber: OK, the DEFAULT_U placement makes sense.

Regarding `constexpr` with variables that are initialized to a constant value, I don't think that it makes a difference, especially since this is `static` and so only done once (though only one instance of this class is ever made, so it's a distinction without a difference). I vote for using `const` to avoid confusing people.

Simulator/Edges/Neuro/AllSpikingSynapses.cpp
```diff
@@ -173,7 +173,7 @@ void AllSpikingSynapses::createEdge(BGSIZE iEdg, int srcVertex, int destVertex,
    sourceVertexIndex_[iEdg] = srcVertex;
    W_[iEdg] = edgSign(type) * 10.0e-9;
    type_[iEdg] = type;
-   tau_[iEdg] = DEFAULT_tau;
+   tau_[iEdg] = AllSpikingSynapses::DEFAULT_tau;
```
stiber, line 176: I don't think this needs the scope specification here.

PR #937, opened 2026-05-14
[Issue 528] Make cluster point process code more general by fixing hard coded values

### LINKED ISSUE #528: Make cluster point process code more general by fixing hard coded values
Currently, the cluster point process script, used to generate input files for the 911 models, has many "magic numbers" and hard-coded parameters for a specific simulation. It needs to be refactored to be more general, where parameters tuned for the simulation are provided when the script is executed.

CodersRepo:
…ph path

Closes #528 

#### Description
Cluster point process: configurable params, JSON CLI, fix ignored graph path

Use the graph file chosen in the UI instead of a hardcoded spd.graphml path.
Add generate_cluster_point_process_xml() for shared GUI/CLI generation with
optional RNG seed, prototype selection weights, and XML clock attributes.

Extend secprocess() with optional prototype_weights; keep legacy 40/50/9/1
weights for four prototypes when weights are omitted; factor Tukey fence
constant. Add --config for headless runs from JSON; validate graph path and
fix mismatched error labels.

- [x] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/Tools/InputGeneration/ClusterPointProcess/cluster_point_process.py
+++ b/Tools/InputGeneration/ClusterPointProcess/cluster_point_process.py
@@ -75,7 +75,7 @@ def generate_cluster_point_process_xml(
 
     Parameters mirror the GUI fields. prototype_weights maps each prototype key to a
     relative frequency (need not sum to exactly 1; values are normalized). If None,
-    legacy 40/50/9/1%% weights apply when there are exactly four prototypes; otherwise
+    legacy 40/50/9/1% weights apply when there are exactly four prototypes; otherwise
     selection is uniform.
 
     random_seed: if None or empty string, the RNG is not re-seeded.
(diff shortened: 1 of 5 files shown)


stiber:
There are a lot of changes here, and while I could just approve it, I think we need to take advantage of this as a opportunity to add in testing so we can be confident that it produces good output. Here are my suggestions:

- First of all, maybe we can go over this together so I'm convinced it works properly.
- Then, from that experience, let's figure out what kind of testing would make sense in the future. Presumably, this would be something like unit and integration tests, but maybe just a set of tests without distinction.
- Then, let's create an issue for generating a new GitHub action to perform tests of the `Tools` whenever any file in there is modified.

Tools/InputGeneration/ClusterPointProcess/cluster_point_process.py
stiber, line 1: Two items:

1. Add a command line option (and function) to create a boilerplate JSON config file and exit the program.
2. Write up a brief README.md that documents the program and its command line (and config parameters). If you aren't familiar enough with the details of the parameters, then just make an issue to add that to the README.

stiber (changes requested):

stiber (approved):

PR #939, opened 2026-05-19
[issue-722] Fix default recorder variable setup

lnanduri2005:
## Summary

This PR fixes gaps in the default recorder setup for issue #722. The main problem was that some recorder variables were either not registered by default or were registered but not actually being filled with useful data before output was saved.

Changes include:
- populating STDP/static edge recorder data from active edges each epoch,
- registering growth variables as dynamic recorder outputs,
- adding NG911 layout coordinates as constant recorder outputs,
- updating recorder documentation to match the current generic recorder path.

## What changed

For STDP/static edge recording, `ConnStatic::updateConnections()` now fills the recorder vectors for edge weight, source vertex, and destination vertex. It pushes one value per active edge into three parallel vectors:

- `WCurrentEpoch_`
- `sourceVertexIndexCurrentEpoch_`
- `destVertexIndexCurrentEpoch_`

For the same index, those three values describe the same active edge for that epoch.

For growth simulations, this PR registers the existing growth values `radii`, `rates`, `outgrowth`, and `deltaR` as dynamic recorder variables.

For NG911, `Layout911` already had `xloc_` and `yloc_`, but those are `DeviceVector<BGFLOAT>` values and cannot be registered directly with the recorder. This PR adds `xlocRecorder_` and `ylocRecorder_` as recorder-compatible mirrors. These do not replace the original layout coordinates; they only make the coordinates available to the recorder.

The `.assign(numVertices_, 0)` calls size those recorder mirror vectors to one value per vertex before they are filled from the GraphML coordinates.

## Scope

This PR keeps the fix focused. It does not:
- modify Hdf5Recorder,
- revive or use Xml911Recorder,
- record CompleteMatrix fields,
- change the broader recorder architecture,
- add a scalar recorder abstraction.

## Documentation

Updated `docs/Notes/RecordersNotes.md` to make the issue #722 recorder behavior easier to follow, including active-edge output, growth dynamic values, NG911 coordinate recording, and Xml911Recorder’s legacy status.

Updated the NG911 user config example to use `XmlRecorder`.

## Verification

- `git diff --check`
- `cmake --build build`
- `cd build && ./tests` — 133 tests passed

Related to #722


lnanduri2005:
I added a `USE_GPU`-guarded `copyEdgeDeviceToHost()` call before reading the edge arrays in `ConnStatic::updateConnections()`. That should make GPU runs record the latest host-side edge state instead of possibly stale values.

lnanduri2005:
I looked into the regression failure starting with test-tiny, since that was the smallest case to check first.

For test-tiny, the generated output is larger because it now includes the four ConnGrowth matrices this PR adds: radii, rates, outgrowth, and deltaR. The current GoodOutput file does not have those yet. The size changes from 468 bytes to 955 bytes, so the added growth output is +487 bytes for this tiny case.

I also generated the NG911 test-small-911 output to make sure the growth matrices were not showing up there. They are not present in the NG911 output. That file changes from 99,043 bytes to 99,407 bytes, so +364 bytes, which appears to be from the expected coordinate output change.

So from what I can tell, the CI failure is from the new growth recorder output not being in the current GoodOutput baselines yet. I wanted to check this first before updating any baselines, since the remaining question is whether these growth variables should be accepted as default regression output or handled differently.

PR #945, opened 2026-06-13
[ISSUE 378] Move Recorder allocation to setup() and delegate via OperationManager.

### LINKED ISSUE #378: Move Recorder memory allocation to setup() method
@jardiamj found out that the Recorder classes allocate memory on initialization. We should probably move memory allocation to a separate setup() to keep the naming consistent. Once this is done, I believe we can delegate the setup step to the OperationsManager.

CodersRepo:
Renames init() to setup() on Recorder implementations so memory allocation follows the same lifecycle pattern as Layout and Connections.

Closes #378 

#### Description
enamed Recorder::init() to setup() so Recorders follow the same lifecycle as Layout and Connections: constructors only register operations, and file/output allocation happens in setup().

XmlRecorder and Hdf5Recorder now register Operations::setup with OperationManager, and Model::setupSim() calls that instead of recorder_->init() directly. Unit tests were updated to match.

- [ ] Added Documentation 
- [ ] Added Unit Tests 

### DIFF
--- a/Simulator/Core/OperationManager.cpp
+++ b/Simulator/Core/OperationManager.cpp
@@ -91,6 +91,8 @@ string OperationManager::operationToString(const Operations &operation) const
          return "printParameters";
       case Operations::loadParameters:
          return "loadParameters";
+      case Operations::setup:
+         return "setup";
       case Operations::serialize:
          return "serialize";
       case Operations::deserialize:
--- a/Simulator/Core/Model.cpp
+++ b/Simulator/Core/Model.cpp
@@ -16,6 +16,7 @@
 #include "Model.h"
 #include "Connections.h"
 #include "Factory.h"
+#include "OperationManager.h"
 #include "ParameterManager.h"
 #include "Recorder.h"
 #include "Simulator.h"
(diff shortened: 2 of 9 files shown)


Simulator/Core/OperationManager.cpp
```diff
@@ -91,6 +91,8 @@ string OperationManager::operationToString(const Operations &operation) const
          return "printParameters";
       case Operations::loadParameters:
          return "loadParameters";
+      case Operations::setup:
+         return "setup";
```
stiber, line 95: Am I wrong, or does no class register a "setup" operation with the `OperationManager` yet? I see three possibilities here:
1. Stick with `Model::setupSim()` and remove "setup" from `OperationManager`.
2. Just stay with "setup" in `OperationManager` and transition to each class registering a setup method.
3. (I got interrupted and lost my train of thought.)

Note that, if different classes' setup methods need to be call in particular order, even if it is a partial order, then `OperationManager` is _not_ the correct approach, as it doesn't guarantee that methods are called in any particular order.

stiber (changes requested):
Regardless of the decision, we need to make one and then, if choice 1 is decided, incorporate that change into `Model`. If 2 or 3 is decided, we can work out what must be done, document that with notes, and create a new issue for changing all setup methods.

PR #949, opened 2026-07-17
[issue-779] Populate STDP edge recorder output

### LINKED ISSUE #779: Ensure Recordable Class Supports Variables for STDP Simulations
In particular, the following edge information needs to be savable:

- source vertex
- destination vertex
- edge weight

@priyadhanu14 is hacking code to do this for growth simulations.

lnanduri2005:
Closes #779

#### Description

This change adds support for recording the source vertex, destination vertex, and weight of active edges during STDP simulations.

I added getters in `AllEdges` so `ConnStatic` can access the edge information. I also added `ConnStatic::updateConnections()` to copy the active edge values into the recorder vectors during each epoch.

I built the project and all 133 tests passed. I also ran `test-tiny-xml-stdp.xml`, which has 4 edges and runs for 3 epochs. The output included `sourceVertex`, `destinationVertex`, and `weight`, with 12 values in each matrix.

- [ ] Added Documentation
- [ ] Added Unit Tests

### DIFF
--- a/Simulator/Connections/Neuro/ConnStatic.cpp
+++ b/Simulator/Connections/Neuro/ConnStatic.cpp
@@ -93,11 +93,31 @@ void ConnStatic::printParameters() const
 void ConnStatic::registerHistoryVariables()
 {
    // Register the following variables to be recorded
-   // Note: There may be potential duplicate weight, source, destination vertices
    Recorder &recorder = Simulator::getInstance().getModel().getRecorder();
    recorder.registerVariable("weight", WCurrentEpoch_, Recorder::UpdatedType::DYNAMIC);
    recorder.registerVariable("sourceVertex", sourceVertexIndexCurrentEpoch_,
                              Recorder::UpdatedType::DYNAMIC);
    recorder.registerVariable("destinationVertex", destVertexIndexCurrentEpoch_,
                              Recorder::UpdatedType::DYNAMIC);
+}
+
+bool ConnStatic::updateConnections()
+{
+   AllEdges &edges = getEdges();
+   const vector<unsigned char> &inUse = edges.getInUse();
+   const vector<BGFLOAT> &weights = edges.getWeights();
+   const vector<int> &sourceVertices = edges.getSourceVertexIndices();
+   const vector<int> &destVertices = edges.getDestVertexIndices();
+
+   // Copy one value per active edge into parallel recorder vectors.
+   // weight/source/destination entries at the same index describe the same edge.
+   for (BGSIZE iEdg = 0; iEdg < inUse.size(); iEdg++) {
+      if (inUse[iEdg]) {
+         WCurrentEpoch_.push_back(weights[iEdg]);
+         sourceVertexIndexCurrentEpoch_.push_back(sourceVertices[iEdg]);
+         destVertexIndexCurrentEpoch_.push_back(destVertices[iEdg]);
+      }
+   }
+
+   return false;
 }
--- a/Simulator/Connections/Neuro/ConnStatic.cpp
+++ b/Simulator/Connections/Neuro/ConnStatic.cpp
@@ -103,6 +103,11 @@ void ConnStatic::registerHistoryVariables()
 
 bool ConnStatic::updateConnections()
 {
+   // GPU STDP is not implemented, so this only uses the CPU edge data.
+   WCurrentEpoch_.startNewEpoch();
+   sourceVertexIndexCurrentEpoch_.startNewEpoch();
+   destVertexIndexCurrentEpoch_.startNewEpoch();
+
    AllEdges &edges = getEdges();
    const vector<unsigned char> &inUse = edges.getInUse();
    const vector<BGFLOAT> &weights = edges.getWeights();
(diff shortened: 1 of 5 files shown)


Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -93,11 +93,31 @@ void ConnStatic::printParameters() const
...
    Recorder &recorder = Simulator::getInstance().getModel().getRecorder();
    recorder.registerVariable("weight", WCurrentEpoch_, Recorder::UpdatedType::DYNAMIC);
    recorder.registerVariable("sourceVertex", sourceVertexIndexCurrentEpoch_,
                              Recorder::UpdatedType::DYNAMIC);
    recorder.registerVariable("destinationVertex", destVertexIndexCurrentEpoch_,
                              Recorder::UpdatedType::DYNAMIC);
 }
+
+bool ConnStatic::updateConnections()
+{
```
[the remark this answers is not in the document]
stiber: In particular, see my questions attached to the `UpdateConnections()` method. A unit test should be able to verify that the `ConnStatic` data members are indeed copies of those from the Edge class.

Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -93,11 +93,31 @@ void ConnStatic::printParameters() const
...
+   for (BGSIZE iEdg = 0; iEdg < inUse.size(); iEdg++) {
+      if (inUse[iEdg]) {
+         WCurrentEpoch_.push_back(weights[iEdg]);
+         sourceVertexIndexCurrentEpoch_.push_back(sourceVertices[iEdg]);
+         destVertexIndexCurrentEpoch_.push_back(destVertices[iEdg]);
+      }
+   }
+
+   return false;
+}
```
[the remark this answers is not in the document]
stiber: Add a comment to indicate that this doesn't ensure data consistency between GPU and CPU. Currently, there is no GPU STDP implementation.

Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -93,11 +93,31 @@ void ConnStatic::printParameters() const
...
+   const vector<int> &sourceVertices = edges.getSourceVertexIndices();
+   const vector<int> &destVertices = edges.getDestVertexIndices();
+
+   // Copy one value per active edge into parallel recorder vectors.
+   // weight/source/destination entries at the same index describe the same edge.
+   for (BGSIZE iEdg = 0; iEdg < inUse.size(); iEdg++) {
+      if (inUse[iEdg]) {
+         WCurrentEpoch_.push_back(weights[iEdg]);
+         sourceVertexIndexCurrentEpoch_.push_back(sourceVertices[iEdg]);
+         destVertexIndexCurrentEpoch_.push_back(destVertices[iEdg]);
```
stiber, line 123: Two questions:

1. Why not record the originals in the Edge class?
2. Alternatively, if it is desirable to copy them to `ConnStatic` each epoch, I think it will be necessary to first clear the three `ConnStatic` data members before starting the copy.

Use the CoPilot unit test skill to generate unit tests for this. It should catch the case where these aren't cleared and so aren't copies. Actually, can't you just use the assignment operator, `std::copy()`, or `vector::assign()?

stiber (changes requested):

lnanduri2005:
I updated this based on your review comments. I reset the three recorder vectors before copying the active edge values, added the CPU-only comment since GPU STDP is not implemented, and added a focused ConnStatic unit test that checks the source, destination, and weight values. The test also calls updateConnections() twice to make sure the values do not keep accumulating. All 134 tests are passing now.

stiber:
Great! Independent of my review comments (if any), make sure then to resolve any previous conversations in this PR that need to be resolved (there's a button for that).

Simulator/Connections/Neuro/ConnStatic.cpp
```diff
@@ -93,11 +93,36 @@ void ConnStatic::printParameters() const
...
    recorder.registerVariable("destinationVertex", destVertexIndexCurrentEpoch_,
                              Recorder::UpdatedType::DYNAMIC);
 }
+
+bool ConnStatic::updateConnections()
+{
+   // GPU STDP is not implemented, so this only uses the CPU edge data.
+   WCurrentEpoch_.startNewEpoch();
+   sourceVertexIndexCurrentEpoch_.startNewEpoch();
+   destVertexIndexCurrentEpoch_.startNewEpoch();
```
stiber, line 109: Right now, these are called in the `Recorder` classes, in their `compileHistories()` methods (at the end, after their contents from the last epoch has been captured). I'm a little hesitant to move this functionality here. Moreover, since that in principle has already been done, I'm a little worried that this is actually masking a bug elsewhere. Ideally, the previous call to `compileHistories()` should have left these in a good state already; if not, I think the fix should be elsewhere.

Simulator/Connections/Neuro/ConnStatic.cpp
stiber: Should reply and resolve this, as appropriate, to finalize this review.

stiber (changes requested):
Hopefully, just some final cleanup. One substantial, new comment.

